PR #1, opened 2018-09-09
Initial bundle setup & feature implementation

weaverryan:
Hi guys!

This adds all the low-level details (composer, travis, etc) + the features themselves, which the README describes. Notes:

* Bundle supports PHP 7.1 and above

**Why this bundle?**

An optional feature in Webpack 4 allows Webpack to "split" an entry (e.g. `app.js`) into multiple, smaller files for performance. But, this means that you need multiple `<script>` tags for each entry... but these are dynamically created and changing. To help with that, Encore dumps a new `entrypoints.json` file, which describes which JS (or CSS) files are needed for each entry. This bundle adds a feature to read those. That's described in the README below (I'll likely move most of the README into the official docs).

Cheers!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+composer.lock
+phpunit.xml
+vendor/
+php_cs.cache
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,0 +1,21 @@
+<?php
+
+if (!file_exists(__DIR__.'/src')) {
+    exit(0);
+}
+
+$finder = PhpCsFixer\Finder::create()
+    ->in(__DIR__.'/src')
+;
+
+return PhpCsFixer\Config::create()
+    ->setRules(array(
+        '@Symfony' => true,
+        '@Symfony:risky' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'protected_to_private' => false,
+        'semicolon_after_instruction' => false,
+    ))
+    ->setRiskyAllowed(true)
+    ->setFinder($finder)
+;
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,43 @@
+language: php
+sudo: false
+cache:
+    directories:
+        - $HOME/.composer/cache/files
+        - $HOME/symfony-bridge/.phpunit
+
+env:
+    global:
+        - PHPUNIT_FLAGS="-v"
+        - SYMFONY_PHPUNIT_DIR="$HOME/symfony-bridge/.phpunit"
+
+matrix:
+    fast_finish: true
+    include:
+        - php: 7.1
+        - php: 7.1
+          env: deps=high
+        - php: 7.2
+          env: deps=low
+
+    allow_failures:
+          # Dev-master is allowed to fail.
+        - env: STABILITY="dev"
+
+before_install:
+    - if [[ $COVERAGE != true ]]; then phpenv config-rm xdebug.ini || true; fi
+    - if ! [ -z "$STABILITY" ]; then composer config minimum-stability ${STABILITY}; fi;
+    - if ! [ -v "$DEPENDENCIES" ]; then composer require --no-update ${DEPENDENCIES}; fi;
+    - composer install
+
+install:
+    - |
+      # Install symfony/flex
+      if [[ $deps = low ]]; then
+          export SYMFONY_REQUIRE='>=2.3'
+      fi
+      composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-master
+    - ./vendor/bin/simple-phpunit install
+
+script:
+    - composer validate --strict --no-check-lock
+    - ./vendor/bin/simple-phpunit $PHPUNIT_FLAGS
--- a/README.md
+++ b/README.md
@@ -1,0 +1,81 @@
+# WebpackEncoreBundle: Symfony integration with Webpack Encore!
+
+This bundle allows you to use the `splitEntryChunks()` feature
+from [Webpack Encore](https://symfony.com/doc/current/frontend.html)
+by reading an `entrypoints.json` file and helping you render all of
+the dynamic `script` and `link` tags needed.
+
+Install the bundle with:
+
+```
+composer require symfony/webpack-encore-bundle
+```
+
+## Configuration
+
+If you're using Symfony Flex, you're done! The recipe will
+pre-configure everything you need in the `config/packages/webpack_encore.yaml`
+file:
+
+```yaml
+# config/packages/webpack_encore.yaml
+webpack_encore:
+    # The path where Encore is building the assets - i.e. Encore.setOutputPath()
+    output_path: '%kernel.public_dir%/build'
+```
+
+## Usage
+
+First, enable the "Split Chunks" functionality in Webpack Encore:
+
+```diff
+// webpack.config.js
+// ...
+    .setOutputPath('public/build/')
+    .setPublicPath('/build')
+    .setManifestKeyPrefix('build/')
+
+    .addEntry('entry1', './assets/some_file.js')
+
++   .splitEntryChunks()
+// ...
+```
+
+When you enable `splitEntryChunks()`, instead of just needing 1 script tag
+for `entry1.js` and 1 link tag for `entry1.css`, you may now need *multiple*
+script and link tags. This is because Webpack ["splits" your files](https://webpack.js.org/plugins/split-chunks-plugin/)
+into smaller pieces for greater optimization. 
+
+To help with this, Encore writes a `entrypoints.json` file that contains
+all of the files needed for each "entry".
+
+For example, to render all of the `script` and `link` tags for a specific
+"entry" (e.g. `entry1`), you can:
+
+```twig
+{# any template or base layout where you need to include a JavaScript entry #}
+
+{% block javascripts %}
+    {{ parent() }}
+
+    {{ encore_entry_script_tags('entry1') }}
+{% endblock %}
+
+{% block stylesheets %}
+    {{ parent() }}
+
+    {{ encore_entry_link_tags('entry1') }}
+{% endblock %}
+```
+
+Assuming that `entry1` required two files to be included - `build/vendor~entry1~entry2.js`
+and `build/entry1.js`, then `encore_entry_script_tags()` is equivalent to:
+
+```twig
+<script src="{{ asset('build/vendor~entry1~entry2.js') }}"></script>
+<script src="{{ asset('build/entry1.js') }}"></script>
+```
+
+If you want more control, you can use the `encore_entry_js_files()` and
+`encore_entry_css_files()` methods to get the list of files needed, then
+loop and create the `script` and `link` tags manually.
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,36 @@
+{
+    "name": "symfony/webpack-encore-bundle",
+    "description": "Integration with your Symfony app & Webpack Encore!",
+    "type": "symfony-bundle",
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Symfony Community",
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "autoload": {
+        "psr-4": {
+            "Symfony\\WebpackEncoreBundle\\": "src"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Symfony\\WebpackEncoreBundle\\Tests\\": "tests/"
+        }
+    },
+    "require": {
+        "php": "^7.1.3",
+        "symfony/config": "^3.4 || ^4.0",
+        "symfony/dependency-injection": "^3.4 || ^4.0",
+        "symfony/http-kernel": "^3.4 || ^4.0"
+    },
+    "require-dev": {
+        "friendsofphp/php-cs-fixer": "^2.13",
+        "symfony/asset": "^3.4 || ^4.0",
+        "symfony/framework-bundle": "^3.4 || ^4.0",
+        "symfony/phpunit-bridge": "^3.4 || ^4.1",
+        "symfony/twig-bundle": "^3.4 || ^4.0",
+        "twig/twig": "^1.35 || ^2.0"
+    }
+}
--- a/src/Asset/EntrypointLookup.php
+++ b/src/Asset/EntrypointLookup.php
@@ -1,0 +1,100 @@
+<?php
+
+/*
+ * This file is part of the Symfony MakerBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\WebpackEncoreBundle\Asset;
+
+/**
+ * Returns the CSS or JavaScript files needed for a Webpack entry.
+ *
+ * This reads a JSON file with the format of Webpack Encore's entrypoints.json file.
+ *
+ * @final
+ */
+class EntrypointLookup
+{
+    private $entrypointJsonPath;
+
+    private $entriesData;
+
+    private $returnedFiles = [];
+
+    public function __construct(string $entrypointJsonPath)
+    {
+        $this->entrypointJsonPath = $entrypointJsonPath;
+    }
+
+    public function getJavaScriptFiles(string $entryName): array
+    {
+        return $this->getEntryFiles($entryName, 'js');
+    }
+
+    public function getCssFiles(string $entryName): array
+    {
+        return $this->getEntryFiles($entryName, 'css');
+    }
+
+    /**
+     * Resets the state of this service.
+     */
+    public function reset()
+    {
+        $this->returnedFiles = [];
+    }
+
+    private function getEntryFiles(string $entryName, string $key): array
+    {
+        $this->validateEntryName($entryName);
+        $entriesData = $this->getEntriesData();
+        $entryData = $entriesData[$entryName];
+
+        if (!isset($entryData[$key])) {
+            throw new \InvalidArgumentException(sprintf('Could not find "%s" key for the "%s" entry.', $key, $entryName));
+        }
+
+        // make sure to not return the same file multiple times
+        $entryFiles = $entryData[$key];
+        $newFiles = array_values(array_diff($entryFiles, $this->returnedFiles));
+        $this->returnedFiles = array_merge($this->returnedFiles, $newFiles);
+
+        return $newFiles;
+    }
+
+    private function validateEntryName(string $entryName)
+    {
+        $entriesData = $this->getEntriesData();
+        if (!isset($entriesData[$entryName])) {
+            $withoutExtension = substr($entryName, 0, strrpos($entryName, '.'));
+
+            if (isset($entriesData[$withoutExtension])) {
+                throw new \InvalidArgumentException(sprintf('Could not find the entry "%s". Try "%s" instead (without the extension).', $entryName, $withoutExtension));
+            }
+
+            throw new \InvalidArgumentException(sprintf('Could not find the entry "%s" in "%s". Found: %s.', $entryName, $this->entrypointJsonPath, implode(', ', array_keys($entriesData))));
+        }
+    }
+
+    private function getEntriesData(): array
+    {
+        if (null === $this->entriesData) {
+            if (!file_exists($this->entrypointJsonPath)) {
+                throw new \InvalidArgumentException(sprintf('Could not find the entrypoints file from Webpack: the file "%s" does not exist.', $this->entrypointJsonPath));
+            }
+
+            $this->entriesData = json_decode(file_get_contents($this->entrypointJsonPath), true);
+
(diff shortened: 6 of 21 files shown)


composer.json
```diff
@@ -0,0 +1,35 @@
...
+            "Symfony\\WebpackEncoreBundle\\": "src"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Symfony\\WebpackEncoreBundle\\Tests\\": "tests/"
+        }
+    },
+    "require": {
+        "php": ">=7.1.3",
```
Nyholm: We do not support PHP8, lets do `^7.1.3` (even tough PHP is not following Semver)

src/Asset/EntrypointLookup.php
```diff
@@ -0,0 +1,91 @@
...
+ * This reads a JSON file with the format of Webpack Encore's entrypoints.json file.
+ * @final
+ */
+class EntrypointLookup
+{
+    private $entrypointJsonPath;
+
+    private $entriesData;
+
+    private $returnedFiles = [];
```
Nyholm, line 27: `array()`

Nyholm: Or, we might be okey with `[]` since fabbot is not active here.

src/Asset/ManifestLookup.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
```
Nyholm, line 2: Should we declare classes as strict?

Nyholm (approved):
I think it looks good. Just found some minors

.gitignore
```diff
@@ -0,0 +1,2 @@
+/vendor/
```
nicolas-grekas: composer.lock
phpunit.xml
vendor/

src/Asset/ManifestLookup.php
weaverryan: I'm fine with that, but we're not using them in `symfony/symfony` yet - I'm not sure if that's a policy, or for other reasons (that don't apply here)

.travis.yml
```diff
@@ -0,0 +1,58 @@
...
+          # Minimum supported dependencies with the latest and oldest PHP version
+        - php: 7.2
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="weak_vendors"
+        - php: 7.0
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="weak_vendors"
+
+          # Test the latest stable release
+        - php: 5.5
+        - php: 5.6
+        - php: 7.0
```
nicolas-grekas: not needed, minimum php is 7.1, isn't it?

.travis.yml
```diff
@@ -0,0 +1,58 @@
...
+        - php: 7.1
+        - php: 7.2
+          env: COVERAGE=true PHPUNIT_FLAGS="-v --coverage-text"
+
+          # Test LTS versions. This makes sure we do not use Symfony packages with version greater
+          # than 3. Read more at https://github.com/symfony/lts
+        - php: 7.2
+          env: DEPENDENCIES="symfony/lts:^3"
```
nicolas-grekas: going to be deprecated, better use global flex + SYMFONY_REQUIRE (see symfony's .travis.yml)

src/Asset/EntrypointLookup.php
weaverryan: Yea, not sure. We use `[]` on MakerBundle, but were forced to not use fabbot due to this :/. Using fabbot is ideal.

nicolas-grekas (commented):
random comments :)

src/Asset/EntrypointLookup.php
nicolas-grekas: no reason to use array() here, as there is no merge conflict to save us from

.travis.yml
Nyholm: I should update our docs for "best practices for reusable bundles"

src/Asset/ManifestLookup.php
nicolas-grekas: Personally, I don't see the benefit of strict mode: it provides no benefit for the "public" facing APIs.
Of course strict mode can help spot some errors inside the boundaries of the facing API, but anyway, this is covered by community work+tests.

.travis.yml
weaverryan: There's a LOT of custom code in symfony's `.travis.yml` file - I'm not sure what's needed. I see the install flex part - https://github.com/symfony/symfony/blob/master/.travis.yml#L198-L205 - I need that, makes sense. But what about the big "running tests" block? https://github.com/symfony/symfony/blob/master/.travis.yml#L216-L244

README.md
```diff
@@ -0,0 +1,85 @@
...
+
+```yaml
+# config/packages/webpack_encore.yaml
+webpack_encore:
+    # The path where Encore is building the assets - i.e. Encore.setOutputPath()
+    output_path: '%kernel.public_dir%/build'
+
+    # The public prefix to your assets that you normally use with the asset() function (e.g. build/) -
+    # should match the "setManifestKeyPrefix()" value in webpack.config.js, if set.
+    asset_path_prefix: 'build/'
```
pierredup: Not sure if it still needs to be added, but this option isn't supported with the current configuration

.gitignore
```diff
@@ -0,0 +1,3 @@
+composer.lock
+phpunit.xml
+vendor/
```
nicolas-grekas: missing EOL (there are more in the PR)

.travis.yml
```diff
@@ -0,0 +1,55 @@
...
+        - PHPUNIT_FLAGS="-v"
+        - SYMFONY_PHPUNIT_DIR="$HOME/symfony-bridge/.phpunit"
+
+matrix:
+    fast_finish: true
+    include:
+          # Minimum supported dependencies with the latest and oldest PHP version
+        - php: 7.2
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="weak_vendors"
+        - php: 7.0
```
nicolas-grekas: not needed

.travis.yml
nicolas-grekas: Yes, only L198-L205, the rest is too specific

src/Asset/ManifestLookup.php
```diff
@@ -0,0 +1,46 @@
...
+                throw new \RuntimeException(sprintf('Asset manifest file "%s" does not exist.', $this->manifestPath));
+            }
+
+            $this->manifestData = json_decode(file_get_contents($this->manifestPath), true);
+            if (0 < json_last_error()) {
+                throw new \RuntimeException(sprintf('Error parsing JSON from asset manifest file "%s" - %s', $this->manifestPath, json_last_error_msg()));
+            }
+        }
+
+        return isset($this->manifestData[$path]) ? $this->manifestData[$path] : null;
```
chalasr: `return $this->manifestData[$path] ?? null;`

composer.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "symfony/webpack-encore-bundle",
+    "description": "Integration with your Symfony app & Webpack Encore!",
+    "type": "symfony-bundle",
+    "license": "MIT",
+    "authors": [
+        {
```
chalasr, line 7: You could add yourself, that's deserved IMHO

README.md
```diff
@@ -0,0 +1,85 @@
+# WebpackEncoreBundle: Symfony integration with Webpack Encore!
+
+This bundle allows you to use the `splitEntryChunks()` feature
+from [Webpack Encore](https://symfony.com/doc/current/frontend.html)
+by reading an `entrypoins.json` file and helping you render all of
```
stof: typo in the file name

src/Asset/EntrypointLookup.php
```diff
@@ -0,0 +1,91 @@
...
+        $this->returnedFiles = array_merge($this->returnedFiles, $newFiles);
+
+        return $newFiles;
+    }
+
+    private function validateEntryName(string $entryName)
+    {
+        $entriesData = $this->getEntriesData();
+        if (!isset($entriesData[$entryName])) {
+            $withoutExtension = substr($entryName, 0, strrpos($entryName, '.'));
```
stof, line 74: what happens if there is no dot in `$entryName` ?

src/Asset/EntrypointLookup.php
```diff
@@ -0,0 +1,91 @@
...
+
+            if (isset($entriesData[$withoutExtension])) {
+                throw new \InvalidArgumentException(sprintf('Could not find the entry "%s". Try "%s" instead (without the extension).', $entryName, $withoutExtension));
+            }
+
+            throw new \InvalidArgumentException(sprintf('Could not find the entry "%s" in "%s". Found: %s.', $entryName, $this->entrypointJsonPath, implode(', ', array_keys($entriesData))));
```
stof, line 80: should we suggest all known ones, or only similar alternatives (as we do in some places in Symfony) ?

src/Asset/EntrypointLookup.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private function getEntryFiles(string $entryName, string $key): array
+    {
+        $this->validateEntryName($entryName);
+        $entriesData = $this->getEntriesData();
+        $entryData = $entriesData[$entryName];
+
+        if (!isset($entryData[$key])) {
+            throw new \InvalidArgumentException(sprintf('Could not find "%s" key for the "%s" entry.', $key, $entryName));
```
stof, line 59: if an entry point does not have JS or CSS, should we treat it as an error, or use an empty list ?

src/Asset/TagRenderer.php
```diff
@@ -0,0 +1,76 @@
...
+            );
+        }
+
+        return implode('', $scriptTags);
+    }
+
+    private function getAssetPath(string $assetPath, string $packageName = null): string
+    {
+        if (null === $this->packages) {
+            throw new \Exception('To render the script or link tags, run "composer require symfony/asset".');
```
stof, line 60: All public APIs of this class are reaching this validation. So why is the Packages dependency an optional one in the constructor ?

src/Asset/TagRenderer.php
```diff
@@ -0,0 +1,76 @@
...
+        return implode('', $scriptTags);
+    }
+
+    public function renderWebpackLinkTags(string $entryName, string $packageName = null): string
+    {
+        $scriptTags = [];
+        foreach ($this->entrypointLookup->getCssFiles($entryName) as $filename) {
+            $scriptTags[] = sprintf(
+                '<link rel="stylesheet" href="%s" />',
+                $this->getAssetPath($filename, $packageName)
```
stof: this is missing HTML escaping.

src/Asset/TagRenderer.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->packages = $packages;
+    }
+
+    public function renderWebpackScriptTags(string $entryName, string $packageName = null): string
+    {
+        $scriptTags = [];
+        foreach ($this->entrypointLookup->getJavaScriptFiles($entryName) as $filename) {
+            $scriptTags[] = sprintf(
+                '<script src="%s"></script>',
+                $this->getAssetPath($filename, $packageName)
```
stof: this is missing HTML escaping.

src/Asset/EntrypointLookup.php
stof, line 27: this class needs to be Resettable, to reset this array. Otherwise it will cause issues when rendering multiple responses in the same PHP process (during tests, or when using a php app server rather than php-fpm)

stof (changes requested):

src/Asset/ManifestLookup.php
dkarlovi: @nicolas-grekas 

> it provides no benefit for the "public" facing APIs.

Doesn't it? Does it not trigger type errors if the API is misused from userland?

nicolas-grekas: Nope, not any, see also discussion in https://github.com/symfony/symfony/issues/28423

dkarlovi: You're right, I misunderstood this comment. :see_no_evil: Thanks!

stof: triggering such error is what types are doing, not strict mode (strict mode in *userland* makes type stricter for arguments)

src/Twig/EntryFilesTwigExtension.php
```diff
@@ -0,0 +1,86 @@
...
+
+    public function __construct(ContainerInterface $container)
+    {
+        $this->container = $container;
+    }
+
+    public function getFunctions()
+    {
+        return [
+            new \Twig_SimpleFunction('encore_entry_js_files', [$this, 'getWebpackJsFiles']),
```
yceruto: The Twig 2.x namespace can be used here too.

stof: yeah, it should even be used, making it future-proof (and Twig 1.x also have it)

.gitignore
BackEndTea: How about adding a `.editorconfig` to the project? This would reduce these kinds of errors.

src/Asset/ManifestLookup.php
weaverryan: Is that the same? Does your version not throw a warning about a non-existent key?

src/Asset/EntrypointLookup.php
weaverryan: It seems to be ok - I have some test cases to make sure we fail with the correct exception for simply wrong entry names - e.g. `foo` as well as wrong entry names just because you added the extension - `foo.js`

src/Asset/EntrypointLookup.php
weaverryan: ... I think all. My thinking is that a typo may not be as common as simply entering something that makes no sense (e.g .`build/foo.js`) because you don't fully understand what you're supposed to put here.

src/Asset/EntrypointLookup.php
weaverryan: Good catch. But I just double-checked this: that IS the behavior: if there is no css files or js files for an entry, an empty array is returned (just added a test for this). This is checking if the `js` or `css` key is completely missing from that entry in `entrypoints.json`, which is not possible (so would be some bad situation, or someone using a custom `entrypoints.json` that is not valid.

src/Asset/TagRenderer.php
weaverryan: Good question. And, there may be a better solution. It's so that I can have:

```xml
<argument type="service" id="assets.packages" on-invalid="null" />
```

For this service's definition. I suppose I could remove this service definition entirely in a compiler pass if `assets.packages` is not present. However, this `TagRenderer` service is itself required by the Twig extension. So *it* would need `on-invalid="null"` and to have some code to throw an exception in the correct situation. I'm not sure if one solution is absolutely better than another, or if there's a general standard.

src/Asset/TagRenderer.php
weaverryan: You're absolutely right. But, this is tricky: the asset path needs the equivalent of `html_attr` escaping from Twig. However, this bundle currently does not rely on Twig, and the logic for this escaping is too big to copy (https://github.com/twigphp/Twig/blob/84b2eb541026dbd39590ba80269ae7676b1bfc46/lib/Twig/Extension/Core.php#L1033) and it's not a good idea anyways, since it's security-related.

The only solution I can think of is to couple the bundle to Twig (no huge deal), and either use that `twig_escape_filter` function directly (bit hacky) or render Twig templates inside of `TagRenderer`.

Any other ideas?

src/Asset/EntrypointLookup.php
weaverryan: Good call. I've added it, though we can't rely on the new ResettableInterface from symfony/contracts yet, as it's not released

src/Asset/ManifestLookup.php
chalasr: It does not, that's the same :)

composer.json
```diff
@@ -0,0 +1,36 @@
...
+            "name": "Symfony Community",
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "autoload": {
+        "psr-4": {
+            "Symfony\\WebpackEncoreBundle\\": "src"
```
BackEndTea, line 14: Shouldn't there be a trailing `/` after src?

src/Asset/ManifestLookup.php
```diff
@@ -0,0 +1,47 @@
...
+
+    public function getManifestPath(string $path): ?string
+    {
+        if (null === $this->manifestData) {
+            if (!file_exists($this->manifestPath)) {
+                throw new \RuntimeException(sprintf('Asset manifest file "%s" does not exist.', $this->manifestPath));
+            }
+
+            $this->manifestData = json_decode(file_get_contents($this->manifestPath), true);
+            if (0 < json_last_error()) {
```
BackEndTea: What about making this `JSON_ERROR_NONE !== json_last_error()`? It would make it more clear what is happening

src/Asset/TagRenderer.php
stof: well, at first, you can still apply normal PHP escaping. As you put this inside a quoted HTML attribute, things are fine (most of the escaping performed by `html_attr` is to allow to use it outside quotes, so for attribute names or unquoted attribute values).

Not doing any escaping can lead to broken HTML here

src/Asset/EntrypointLookup.php
stof: well, as this is still a WIP, I think we could delay the stable release of this package until the release of symfony/contracts in November :smile:

composer.json
stof: not needed. Composer would trim it if it exists.

src/Asset/ManifestLookup.php
stof: I agree here

src/Asset/TagRenderer.php
weaverryan: Ah, ok. I've just used `htmlentities` and added a test. If you know of any issues with that, please let me know!

weaverryan:
Updated! In a few days (and after the tests are working), I'll merge this. So, any additional feedback is warmly appreciated :).

src/Asset/TagRenderer.php
stof: or remove also the Twig extension when it does not have the needed dependency, which is what we do in symfony itself.

stof:
This should wait for https://github.com/symfony/webpack-encore/pull/410 which changes a bit the structure of the entrypoints file (CSS becomes a string rather than an array of strings)

.php_cs.dist
```diff
@@ -0,0 +1,38 @@
+<?php
+
+if (!file_exists(__DIR__.'/src')) {
+    exit(0);
+}
+
+$finder = PhpCsFixer\Finder::create()
+    ->in(__DIR__.'/src')
+    // the PHP template files are a bit special
+    ->notName('*.tpl.php')
```
stof: do we have such files ?

.php_cs.dist
```diff
@@ -0,0 +1,38 @@
...
+;
+
+if (\PHP_VERSION_ID < 70100) {
+    // EntityRegenerator works only with PHP 7.1+
+    $finder->notName('EntityRegenerator.php');
```
stof: that looks wrong here. Bad copy-paste ?

.php_cs.dist
```diff
@@ -0,0 +1,38 @@
...
+return PhpCsFixer\Config::create()
+    ->setRules(array(
+        '@Symfony' => true,
+        '@Symfony:risky' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'protected_to_private' => false,
+        'semicolon_after_instruction' => false,
+        'header_comment' => [
+            'header' => <<<EOF
+This file is part of the Symfony MakerBundle package.
```
stof: wrong

simple-phpunit
```diff
@@ -0,0 +1,274 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+// Please update when phpunit needs to be reinstalled with fresh deps:
+// Cache-Id-Version: 2017-11-22 09:30 UTC
+
+error_reporting(-1);
+
+$getEnvVar = function ($name, $default = false) {
```
stof: this file looks wrong. Install the phpunit-bridge instead of duplicating its code.

stof (changes requested):

simple-phpunit
weaverryan: Temporary code - debugging a weird Travis-only phpunit issue :/

Nyholm:
Since this is the first lines of code to repo, I suggest we should merge it now. It may not be perfect but it does not have to be perfect now. 

If this is merged now we all can help to make it better. 

:+1: good job

Issue #4, opened 2018-10-23
Make EntrypointLookup handle missing css key


weaverryan:
In Encore, the `css` key for an asset may be missing entirely. Let's make sure the EntrypointLookup can handle a missing `css` key (just return no assets).

PR #5, opened 2018-10-24
Fix the issue where keys might be missing if no files are found by webpack plugin

codayblue:
Altered the entry point lookup to handle if a key is not found in the entry point file

This will fix issue #4 since we will now return a blank array if the file type is not found. This bug was caused by the swapping of plugins in the encore package that will now handle generating the entry points file. The new plugin will not put in a file type key if there none found.


weaverryan:
Thank you Cody!

Issue #2, opened 2018-10-23
Make EntrypointLookup "changeable"


weaverryan:
In some cases, you may want to use a different implementation of reading the entrypoint data (I've talked with someone about a use-case). So, we should:

1) Create an interface for the lookup
2) Add configuration so a new service can be "swapped" in

jkabat:
Ryan, first of all, thanks for the great package.
I'm not sure if this is the right place to bring it up, but since it relates to entrypoints, I'll ask here. 

I'm using two different webpack configurations (something like https://github.com/symfony/webpack-encore/issues/113 to give you an idea) and because one of them is located in subfolder, Im not able to use twig helpers to generate script and link tags for these entries.

Only the main entrypoints.json is found and used, not the one configured as different package under a subfolder.

My question is if I missed something in documentation or is it still not yet supported feature?

There also may be way to merge entrypoints files, but I'm too new to webpack to be aware of such possibility.

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Could I get a reply or should I close this?

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

PR #6, opened 2018-10-25
Create interfaces to make EntryLookup and ManifestLookup are interchangable

codayblue:
This will fix issue #2 

This is a very simple change to create a set of interfaces to make parts of the bundle changeable by the user. With this change it makes the bundle more flexible because people can other developers can change the functionality to match what they need.


src/Asset/EntrypointLookupInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Symfony\WebpackEncoreBundle\Asset;
```
weaverryan, line 10: Missing license header. Rebase to the current master and run `php vendor/bin/php-cs-fixer fix` and it'll just handle all of this automatically

src/Asset/EntrypointLookupInterface.php
```diff
@@ -0,0 +1,16 @@
...
+interface EntrypointLookupInterface
+{
+    public function getJavaScriptFiles(string $entryName): array;
+
+    public function getCssFiles(string $entryName): array;
+
+    /**
+     * Resets the state of this service.
+     */
+    public function reset();
```
weaverryan, line 29: TODO for later - when Symfony 4.2 is released, we can make this interface extend https://github.com/symfony/contracts/blob/master/Service/ResetInterface.php

src/Asset/EntrypointLookupInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Symfony\WebpackEncoreBundle\Asset;
+
+
+interface EntrypointLookupInterface
+{
+    public function getJavaScriptFiles(string $entryName): array;
```
weaverryan, line 19: Now that I see this interface, I think we should "clean it up" a little bit. I think we should create a new exception class - e.g. `EntrypointNotFoundException` and advertise above this method in phpdoc that this exception should be thrown if the entry does not exist.

src/Asset/ManifestLookupInterface.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Symfony\WebpackEncoreBundle\Asset;
+
+
+interface ManifestLookupInterface
+{
+    public function getManifestPath(string $path): ?string;
```
weaverryan: Same as the EntrypointLookupInterface: let's make this return `string` (not `?string`) and create a new `ManifestKeyNotFoundException` that we should throw if the key does not exist (and rename `$path` to `$key`).

weaverryan (commented):
Hey Cody!

I just "hjacked" this PR a little bit - I think we should make a few tweaks to the interfaces themselves so that they're more solid and specific (this was actually my mistake - but when you created the interfaces, it was more obvious).

Also, we can do it in another PR, but we (at some point) will need to create some new bundle configuration so that the user can specify an alternate service to be used for the entrypoint lookup and manifest lookup.

Thanks!

codayblue:
I made the changes requested. I assumed that the exceptions should be implemented. So I tried implementing the way I thought they were meant to be.

codayblue:
@weaverryan is there anything else needed for this PR?

src/Asset/EntrypointLookup.php
```diff
@@ -31,7 +33,13 @@ public function __construct(string $entrypointJsonPath)
 
     public function getJavaScriptFiles(string $entryName): array
     {
-        return $this->getEntryFiles($entryName, 'js');
+        $results = $this->getEntryFiles($entryName, 'js');
+
+        if (empty($results)) {
+            throw new EntrypointNotFoundException();
```
weaverryan: I think this needs to be moved down to replace the 2 `InvalidArgumentException` in `validateEntryName`. It's actually ok if `$this->getEntryFiles()` returns an empty array: that's possible if you pass a *valid* entry name, but there are no js for it, for example.

This will also make `getCssFiles()` instantly throw this new exception under the same conditions.

src/Asset/EntrypointLookupInterface.php
```diff
@@ -0,0 +1,37 @@
...
+use Symfony\WebpackEncoreBundle\Exception\EntrypointNotFoundException;
+
+interface EntrypointLookupInterface
+{
+    /**
+     * @param string $entryName
+     *
+     * @return array
+     *
+     * @throws EntrypointNotFoundException for if a entry name is passed in and no results were found
```
weaverryan: `@throws EntrypointNotFoundException if an entry name is passed that does not exist in entrypoints.json`

src/Asset/EntrypointLookupInterface.php
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public function getJavaScriptFiles(string $entryName): array;
+
+    /**
+     * @param string $entryName
+     *
+     * @return array
+     */
+    public function getCssFiles(string $entryName): array;
```
weaverryan, line 24: Same `@throws` above this

src/Asset/EntrypointLookupInterface.php
```diff
@@ -0,0 +1,37 @@
...
+     * @return array
+     *
+     * @throws EntrypointNotFoundException for if a entry name is passed in and no results were found
+     */
+    public function getJavaScriptFiles(string $entryName): array;
+
+    /**
+     * @param string $entryName
+     *
+     * @return array
```
weaverryan: No need for the `@param` or `@return` as they are specified via the type-hints / return type

src/Asset/EntrypointLookupInterface.php
```diff
@@ -0,0 +1,37 @@
...
+namespace Symfony\WebpackEncoreBundle\Asset;
+
+use Symfony\WebpackEncoreBundle\Exception\EntrypointNotFoundException;
+
+interface EntrypointLookupInterface
+{
+    /**
+     * @param string $entryName
+     *
+     * @return array
```
weaverryan: No need for the `@param` or `@return` as they are specified via the type-hints / return type

src/Asset/ManifestLookup.php
```diff
@@ -9,12 +9,14 @@
...
 
+use Symfony\WebpackEncoreBundle\Exception\ManifestKeyNotFoundException;
+
 /**
  * Looks up values in Encore's `manifest.json` and returns them.
  *
  * @final
  */
-class ManifestLookup
+class ManifestLookup implements ManifestLookupInterface
```
weaverryan: Sorry - I killed this class entirely

src/Asset/EntrypointLookupInterface.php
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public function getCssFiles(string $entryName): array;
+
+    /**
+     * Resets the state of this service.
+     * TODO for later - when Symfony 4.2 is released, use https://github.com/symfony/contracts/blob/master/Service/ResetInterface.php instead.
```
weaverryan: Don't include this TODO here. Sorry, I realized I confused you on this part! I was just dropping a note in the PR - but we usually don't add things like this in the code.

weaverryan (changes requested):
Ok, one more round of changes! I've caused you some conflicts as you expected (sorry!) and I've deleted the ManifestLookup system entirely, so that will simplify!

Thanks!

codayblue:
@weaverryan I have made the requested changes and resolved the conflicts

weaverryan (approved):

weaverryan:
Thank you Cody!

PR #7, opened 2018-10-26
Add a dependency on symfony/asset

weaverryan:
Hi guys!

I originally avoided a hard dependency on `symfony/asset`, but this adds it? Why:

1) It would allow us to abandon `symfony/webpack-encore-pack`... as we can now attach the recipe to this bundle and it will bring in the needed symfony/asset package

2) Needing to use Encore inside Symfony... but NOT wanting the `symfony/asset` component is a huge edge case. And if you're in that situation, it's trivial to implement your own system for loading `entrypoints.json`.

In short, this simplifies things!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,13 +21,13 @@
     },
     "require": {
         "php": "^7.1.3",
+        "symfony/asset": "^3.4 || ^4.0",
         "symfony/config": "^3.4 || ^4.0",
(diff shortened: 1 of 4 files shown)


codayblue (approved):
We should merge this before #6 so that way I can resolve the conflicts in a couple places.

PR #11, opened 2018-10-30
Read data from an "entrypoints" key instead of the root of the entrypoints.json file

Lyrkan:
This PR changes how the `entrypoints.json` lookup works in order to support https://github.com/symfony/webpack-encore/pull/424.

Currently the entrypoints are directly listed at the root of the `entrypoints.json` file:

```json
{
  "my_entry": {
    "js": [
        "build/file1.js",
        "build/file2.js"
    ],
    "css": [
      "build/styles.css",
      "build/styles2.css"
    ]
  },
  "other_entry": {
    "js": [
        "build/file1.js",
        "build/file3.js"
    ]
  }
}
```

This works fine but will be a bit annoying later on if we want to add additional data such as [files hashes](https://github.com/symfony/webpack-encore/issues/418) that could be used to automatically handle the `integrity` attribute of the generated `<script>` tags.

A solution to that issue would be to move all the current data under an `entrypoints` key and add new keys when needed (for instance `hashes`):

```json
{
  "entrypoints": {
    "my_entry": {
      "js": [
          "build/file1.js",
          "build/file2.js"
      ],
      "css": [
        "build/styles.css",
        "build/styles2.css"
      ]
    },
    "other_entry": {
      "js": [
          "build/file1.js",
          "build/file3.js"
      ]
    }
  }
}
```

### DIFF
--- a/src/Asset/EntrypointLookup.php
+++ b/src/Asset/EntrypointLookup.php
@@ -51,7 +51,7 @@ class EntrypointLookup
     {
         $this->validateEntryName($entryName);
         $entriesData = $this->getEntriesData();
-        $entryData = $entriesData[$entryName];
+        $entryData = $entriesData['entrypoints'][$entryName];
 
         if (!isset($entryData[$key])) {
             // If we don't find the file type then just send back nothing.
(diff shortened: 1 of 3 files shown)


weaverryan:
Thanks @Lyrkan!

Issue #3, opened 2018-10-23
Add a "cached" version of EntrypointLookup in production


weaverryan:
Reference: https://twitter.com/frankdejonge/status/1054367165430800385

codayblue:
@weaverryan do we want to implement a PSR-6 caching version of the classes and let people inject a PSR-6 cache of their choice? I could also go down a route where in the current classes we could see if they injected a cache. If no cache then just move on if it is passed in then use it. Or do we just want something like php file cache so it gets saved into opcache?

weaverryan:
Hmm. Good question. The implementation by Frank (https://gist.github.com/frankdejonge/bba07f9ddb0756d362bed9bc3ffd5e43#file-useprodassetmappass-php) does not involve a cache at all - it just reads the `manifest.json` file at "build" time, and so the contents of `manifest.json` are available in the container.

The advantage is that this is simple :). It *does* require the `manifest.json` file to be available at "build" time... which could be seen as a disadvantage. But... I kinda see that as a *good* thing: you could build your container cache on a "build" server... and then the `manifest.json` file wouldn't even need to be present on the production server.

However... I'm still not totally sure. Let me ask for an opinion - I'd want to follow some patterns in core, as much as possible.

weaverryan:
After chatting, I think we should:

A) Do Frank's approach - with 2 "manifest" readers - a version where we pass it the JSON path and it reads it at runtime (for dev) and another that accepts the paths as an array (and we read the JSON file at container build time).

B) To make the "cached" version used on production, we'll add some new config. I'm thinking:

```yml
webpack_encore:
    cache_entrypoints: true
```

C) When the "cache" option is turned on, if the `entrypoints.json` files doesn't exist, we'll throw an exception.

How does this sound?

codayblue:
Sounds good to me I can start working on this now.

codayblue:
This is being blocked by #2 We need the interface implemented so we can swap out which class to use.

carsonbot:
Thank you for this issue.
There has not been a lot of activity here for a while. Has this been resolved?

carsonbot:
Could I get an answer? If I do not hear anything I will assume this issue is resolved or abandoned. Please get back to me <3

carsonbot:
Hey,

I didn't hear anything so I'm going to close it. Feel free to comment if this is still relevant, I can always reopen!

PR #12, opened 2018-11-04
Created cached entrypoint look up class

codayblue:
- Created the CacheEntrypointLookup
- modified the EntrypointLookup to have a protected method instead of private to easily just extend it
- Created the cache compiler pass that reads the file and will save it to the symfony cache with the service construction
- Created a config option to enable the cache class and compiler pass
- Enabled the compiler pass in the bundle
- Updated tests

This will resolve issue #3


codayblue:
@weaverryan is there anything else needed with this PR?

src/Asset/CacheEntrypointLookup.php
```diff
@@ -0,0 +1,25 @@
...
+
+    public function __construct(array $entrypoints)
+    {
+        $this->entrypoints = $entrypoints;
+    }
+
+    protected function getEntriesData(): array
+    {
+        return $this->entrypoints;
+    }
```
weaverryan: Hmm. Unfortunately, making this protected isn't the best way to make this extension point - though it was the easiest! ;) It forces us to make the method protected... and it also forces us to remove the `@final` on the parent class (which we would like to not need to do).

I see why you did it - there is actually a decent amount of logic in `EntrypointLookup` other than just `json_decode()` on the `manifest.json` file. 

I have a different idea. Sorry for the back-and-forth - this isn't an area that I usually touch in the Symfony core :). Here's my idea:

1) Let's *always* cache. Basically, we will move the caching logic into EntrypointLookup and not make it optional
2) But, we will not use, for example, the `cache.system` cache or something like that. Instead, there is a system in the `Config` component of Symfony for this type of caching - an object called the `ConfigCacheFactory`.

Basically, we use the `ConfigCacheFactory` to cache the results of parsing the `entrypoints.json` file. But, built into that system is a mechanism that will automatically recache it if the file modified time of `entrypoints.json` is updated (in dev only, of course). It means that, in dev, the cache is loaded the first time you access the page, then reloaded whenever the file changes. In `prod`, we would create a new "cache warmer" so that it's loaded during `cache:warmup`.

A good example of this system is the translator: 

* https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Resources/config/translation.xml#L19-L21
* https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Translation/Translator.php#L323-L327

Can you try this out? I think it's ultimately the best solution.

Thanks!

codayblue: I like this solution better. I just did not know about config cache. I will implement this.

src/Asset/EntrypointLookup.php
```diff
@@ -26,9 +30,27 @@ class EntrypointLookup implements EntrypointLookupInterface
...
         $this->entrypointJsonPath = $entrypointJsonPath;
+
+        if (isset($options['debug'])) {
+            $this->debug = $options['debug'];
+        } else {
+            $this->debug = true;
+        }
+
+        if (!isset($options['cache_dir'])) {
+            throw new RuntimeException('Cache directory not set.');
```
weaverryan: If it's not set, then we basically just need to disable caching. This is needed, at the very least, for backwards compatibility - we can't add a required option/argument to this class without a BC layer. However, we *can* trigger a deprecation warning so that we can make this option required in the future.

src/Asset/EntrypointLookup.php
```diff
@@ -89,7 +116,17 @@ private function getEntriesData(): array
...
 
-            $this->entriesData = json_decode(file_get_contents($this->entrypointJsonPath), true);
+            $cache = $this->getConfigCacheFactory()->cache($this->getConfigCacheFileName(), function (ConfigCacheInterface $cache) {
+                $content = sprintf(<<<EOF
+<?php
+return %s;
+EOF
+                , var_export(json_decode(file_get_contents($this->entrypointJsonPath), true), true));
+
+                $cache->write($content);
```
weaverryan: Here, I believe the second argument is `Resource` - we need 1 resource that points to the `entrypoints.json` file. This is the key to how cache invalidation works in the dev environment: by passing that here, the cache will know to regenerate if that file is updated.

weaverryan (commented):
NIce work! I left a few comments - and I know of one other thing we'll need: a cache warmer (so the cache file can be created during `cache:warmup`). For example: https://github.com/symfony/symfony/blob/254f4c8f7f116733fb32cbb872f9184c094630b1/src/Symfony/Bundle/FrameworkBundle/CacheWarmer/TranslationsCacheWarmer.php

codayblue:
@weaverryan I have updated the PR. Sorry for the long wait. Also all the tests passed? There appears to be a deprecation warning from PHP unit that caused Travis to fail.

weaverryan:
I'm going to wait for #17 and #21 before this one - especially #21, which may need to add a constructor arg to `EntrypointLookup` to fix a bug.

Thanks!

src/Asset/EntrypointLookup.php
```diff
@@ -100,6 +122,33 @@ private function getEntriesData(): array
...
+<?php
+return %s;
+EOF
+                , var_export(json_decode(file_get_contents($this->entrypointJsonPath), true), true));
+
+            $cache->write($content, [new FileResource($this->entrypointJsonPath)]);
+        });
+    }
+
+    public function warmUp($cacheDir)
```
weaverryan: The warmup should now use the cache service to read and set the cache - no more ConfigCacheFactory stuff

src/Resources/config/services.xml
```diff
@@ -9,6 +9,9 @@
 
         <service id="webpack_encore.entrypoint_lookup" class="Symfony\WebpackEncoreBundle\Asset\EntrypointLookup">
             <argument /> <!-- entrypoints.json path -->
+            <call method="setCache">
+                <argument key="$cache" type="service" id="webpack_encore.cache" />
+            </call>
```
weaverryan: One tricky thing is that there are now more than just this one service for this class. Look at the Extension class in this bundle - you’re now able to configure multiple builds and each will have its own service. We need to call setCache on all of these AND make the warmer work for all of them.

src/CacheWarmer/EntrypointCacheWarmer.php
```diff
@@ -0,0 +1,50 @@
...
+    public function isOptional()
+    {
+        return true;
+    }
+
+    public function warmUp($cacheDir)
+    {
+        $entryPointCollection = $this->container->get('webpack_encore.entrypoint_lookup_collection');
+
+        if ($entryPointCollection instanceof EntrypointLookupCollection) {
```
weaverryan: We could simplify if we added a getAll() method to this object. Then we wouldn’t need to inject builds here. We could just foreach over getAll()

src/Asset/EntrypointLookupInterface.php
```diff
@@ -9,10 +9,11 @@
 
 namespace Symfony\WebpackEncoreBundle\Asset;
 
+use Symfony\Component\HttpKernel\CacheWarmer\WarmableInterface;
 use Symfony\Contracts\Service\ResetInterface;
 use Symfony\WebpackEncoreBundle\Exception\EntrypointNotFoundException;
 
-interface EntrypointLookupInterface extends ResetInterface
+interface EntrypointLookupInterface extends ResetInterface, WarmableInterface
```
weaverryan: What’s the purpose of the WarmableInterface?

src/CacheWarmer/EntrypointCacheWarmer.php
```diff
@@ -0,0 +1,50 @@
...
+            foreach ($this->builds as $build => $path) {
+                $fullPath = $path.'/entrypoints.json';
+
+                // If the file does not exist then just skip past this entry point.
+                if (!file_exists($fullPath)) {
+                    continue;
+                }
```
weaverryan: I think this logic shouldn’t be necessary. We can just call warmUp() - that method is already written to not fail if this file is missing

src/Resources/config/services.xml
```diff
@@ -32,5 +35,22 @@
...
+            <argument type="service" id="Symfony\Component\DependencyInjection\ContainerInterface" />
+        </service>
+
+        <service id="webpack_encore.cache_fallback" class="Symfony\Component\Cache\Adapter\FilesystemAdapter">
+            <argument />
+            <argument>0</argument>
+            <argument key="$directory">%kernel.cache_dir%/encore/</argument>
+        </service>
+
+        <service id="webpack_encore.cache" class="Symfony\Component\Cache\Adapter\PhpArrayAdapter">
```
weaverryan: I think we’ll need a new config option to opt into this behavior. If that config option is false, I think we calcium’s just remove this service in the Extension class. Then we just need the on-invalid behavior on the setCache calls to be correct so Symfony just skips the call instead of an error

src/CacheWarmer/EntrypointCacheWarmer.php
codayblue: The container interface does not have a get all. Is that a method on the class that gets injected?

src/Asset/EntrypointLookupInterface.php
codayblue: This is to make sure the warmup method is there. I can remove the interface and just make a warmup method in this interface. Another thing is I can remove the interface and not have the warm up method here and the cache warmer can just assume the method exists.

src/CacheWarmer/EntrypointCacheWarmer.php
codayblue: I will remove this.

src/Resources/config/services.xml
codayblue: Do you know what you want the config option to be named?

src/Asset/EntrypointLookupInterface.php
weaverryan: I think we should remove it - the warmer knows exactly what instances it’s workijg with and therefore what methods exist. What’s the purpose of this interface normally?

src/CacheWarmer/EntrypointCacheWarmer.php
weaverryan: Yea, I think we should actually inject the EntrypointLookupCollection service here directly - I think there’s no need to inject the container and then fetch it. Then, yes, the getAll() method would be on *that* class (the collection class) - it will allow us to easily ask for all the EntrypointLookup instances.

src/Resources/config/services.xml
weaverryan: Probably could be cache: true on the “root” level of the bundle config. IMO, no need to make it possible to disable/enable cache independently for each build - someone can ask for that later. Just one global option.

src/Asset/EntrypointLookupInterface.php
codayblue: The warmer knows about the collection and this interface. It does not know about the actual instance. I can have it do an instance check and on what is returned from the collection.

src/CacheWarmer/EntrypointCacheWarmer.php
codayblue: The collection class gets an instance of ContainerInterface this does not have a get all. So should I pass the builds to the collection? So it can track them. When I debug through I see it is the ServiceLocator class from the dependence injection component. I dont see a get all method there but I can see how it works and try to extend it to make that possible.

src/Resources/config/services.xml
```diff
@@ -9,6 +9,9 @@
 
         <service id="webpack_encore.entrypoint_lookup" class="Symfony\WebpackEncoreBundle\Asset\EntrypointLookup">
             <argument /> <!-- entrypoints.json path -->
+            <call method="setCache">
+                <argument key="$cache" type="service" id="webpack_encore.cache" />
```
weaverryan: Remove the `$key` here - the whole "binding by named arguments" thing is really something that should only be used in application land. In a bundle, we should avoid this and pass arguments by order - it'll avoid a minor performance impact when rebuilding the container.

codayblue:
@weaverryan Is there anything else needed for this PR? Are we waiting on another PR?

weaverryan:
Thank you @codayblue and @nicolas-grekas - this was a bit more complex than I gave it credit for initially :)

PR #15, opened 2018-11-11
Don't self-close link tag

Seb33300:
Self-closing is only required by XHTML, invalid in HTML4 and optional in HTML5.
Nowadays, most of application are HTML5 so we should avoid it since it will save 2 chars from the output.


weaverryan (approved):
I meant to do this :).

weaverryan:
Thanks Sébastien!

Issue #13, opened 2018-11-07
Is there a way to work with multiple entrypoints ?


shulard:
Hello !

I just tried to install this new bundle coupled with the webpack-encore v0.21.0 and faced a little issue.

In my project we have multiple Encore configuration which export assets in multiple output path. Maybe it's not the recommended way to use it but with that setup I can't use the new features because every call to twig functions will read the entry point defined in the output path.

Before that, we used the asset package feature to declare multiple packages (each one with it's own manifest.json file).

I checked the Bundle code and find that the asset packages are still used but the entrypoint.json file is always the same.

Is there a way to handle that case ?

B-Galati:
Hello!

FYI I have a project with pretty much the same setup and same issues (I did not try yet)

weaverryan:
Hi @shulard!

You're right that this is not currently supported. No problem - let's figure out how! First, your setup has a different outputPath for each build, correct? So you are already able to generate multiple `entrypoints.json` files?

If so, then the issue is just with the bundle: I think there needs to be a way to define multiple "builds", each with a unique key and different config for the output dir. So, it might look like:

```yaml
webpack_encore:
    # the "default" build
    output_path: '%kernel.public_dir%/build'
    builds:
        frontend: '%kernel.public_dir%/frontend/build'
```

Then in Twig:

```twig
{{ encore_entry_script_tags('entry1', 'frontend') }}

{# or, really, it may need to be the third argument #}
{{ encore_entry_script_tags('entry1', null, 'frontend') }}
```

The tricky part is that `encore_entry_script_tags()` already has a 2nd argument - it's the "asset package name". That's why I listed the other option.

Thoughts?

shulard:
Hello @weaverryan !

Yep I already checked the second argument for `encore_entry_script_tags` but I think that if we use multiple build the asset package feature is not relevant anymore (or in very specific edge cases...).

I generate multiple entrypoints.json files so your solution seems the right one. It's also compatible with the asset logic.

The only remaining questions : Is using the same logic for asset package and encore build a bit confusing for the users ? Is it relevant to support asset packages in encore twig helpers ?

chiqui3d:
This option is necessary, after following this tutorial https://symfony.com/doc/current/frontend/encore/advanced-config.html#defining-multiple-webpack-configurations, I am left without being able to apply a different path for the front and admin

```
webpack_encore:
    # the "default" build
    output_path: 
         front :'%kernel.public_dir%/build/front'
[...]

PR #17, opened 2018-11-15
Allow to define multiple entrypoints and load assets from each of them.

shulard:
This PR is added to fix the issue #13. With that new code, you can configure your project like this : 

```yaml
webpack_encore:
    # the "default" build
    output_path: '%kernel.project_dir%/public/build'
    builds:
        frontend: '%kernel.project_dir%/public/frontend/build'
```

Then in your Twig template : 

```twig
{{ encore_entry_script_tags('entry1', 'frontend') }}
```

### DIFF
--- a/src/Asset/EntrypointLookupCollection.php
+++ b/src/Asset/EntrypointLookupCollection.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Symfony WebpackEncoreBundle package.
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\WebpackEncoreBundle\Asset;
+
+use Symfony\WebpackEncoreBundle\Exception\UndefinedBuildException;
+use Psr\Container\ContainerInterface;
+
+/**
+ * Aggregate the different entry points configured in the container.
+ *
+ * Retrieve the EntrypointLookup instance from the given key.
+ *
+ * @final
+ */
+class EntrypointLookupCollection
+{
+    private $buildEntrypoints;
+
+    public function __construct(ContainerInterface $buildEntrypoints)
+    {
+        $this->buildEntrypoints = $buildEntrypoints;
+    }
+
+    public function getEntrypointLookup(string $buildName): EntrypointLookupInterface
+    {
+        if (!$this->buildEntrypoints->has($buildName)) {
+            throw new UndefinedBuildException(sprintf('Given entry point "%s" is not configured', $buildName));
+        }
+
+        return $this->buildEntrypoints->get($buildName);
+    }
+}
--- a/src/Asset/TagRenderer.php
+++ b/src/Asset/TagRenderer.php
@@ -10,23 +10,39 @@
 namespace Symfony\WebpackEncoreBundle\Asset;
 
 use Symfony\Component\Asset\Packages;
+use Symfony\Component\DependencyInjection\ServiceLocator;
 
 final class TagRenderer
 {
-    private $entrypointLookup;
+    private $entrypointLookupCollection;
 
     private $packages;
 
-    public function __construct(EntrypointLookupInterface $entrypointLookup, Packages $packages)
-    {
-        $this->entrypointLookup = $entrypointLookup;
+    public function __construct(
+        $entrypointLookupCollection,
+        Packages $packages
+    ) {
+        if ($entrypointLookupCollection instanceof EntrypointLookupInterface) {
+            @trigger_error(sprintf('The "$entrypointLookupCollection" argument in method "%s()" must be an instance of EntrypointLookupCollection.', __METHOD__), E_USER_DEPRECATED);
+
+            $this->entrypointLookupCollection = new EntrypointLookupCollection(
+                new ServiceLocator(['_default' => function() use ($entrypointLookupCollection) {
+                    return $entrypointLookupCollection;
+                }])
+            );
+        } elseif ($entrypointLookupCollection instanceof EntrypointLookupCollection) {
+            $this->entrypointLookupCollection = $entrypointLookupCollection;
+        } else {
+            throw new \TypeError('The "$entrypointLookupCollection" argument must be an instance of EntrypointLookupCollection.');
+        }
+
         $this->packages = $packages;
     }
 
-    public function renderWebpackScriptTags(string $entryName, string $packageName = null): string
+    public function renderWebpackScriptTags(string $entryName, string $packageName = null, string $entrypointName = '_default'): string
     {
         $scriptTags = [];
-        foreach ($this->entrypointLookup->getJavaScriptFiles($entryName) as $filename) {
+        foreach ($this->getEntrypointLookup($entrypointName)->getJavaScriptFiles($entryName) as $filename) {
             $scriptTags[] = sprintf(
                 '<script src="%s"></script>',
                 htmlentities($this->getAssetPath($filename, $packageName))
@@ -36,10 +52,10 @@ final class TagRenderer
         return implode('', $scriptTags);
     }
 
-    public function renderWebpackLinkTags(string $entryName, string $packageName = null): string
+    public function renderWebpackLinkTags(string $entryName, string $packageName = null, string $entrypointName = '_default'): string
     {
         $scriptTags = [];
-        foreach ($this->entrypointLookup->getCssFiles($entryName) as $filename) {
+        foreach ($this->getEntrypointLookup($entrypointName)->getCssFiles($entryName) as $filename) {
             $scriptTags[] = sprintf(
                 '<link rel="stylesheet" href="%s">',
                 htmlentities($this->getAssetPath($filename, $packageName))
@@ -59,5 +75,10 @@ final class TagRenderer
             $assetPath,
             $packageName
         );
+    }
+
+    private function getEntrypointLookup(string $buildName): EntrypointLookupInterface
+    {
+        return $this->entrypointLookupCollection->getEntrypointLookup($buildName);
     }
 }
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -12,6 +12,7 @@ namespace Symfony\WebpackEncoreBundle\DependencyInjection;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
+use Symfony\Component\Config\Definition\Exception\InvalidDefinitionException;
 
 final class Configuration implements ConfigurationInterface
 {
@@ -26,6 +27,19 @@ final class Configuration implements ConfigurationInterface
                 ->scalarNode('output_path')
                     ->isRequired()
                     ->info('The path where Encore is building the assets - i.e. Encore.setOutputPath()')
+                ->end()
+                ->arrayNode('builds')
+                    ->useAttributeAsKey('name')
+                    ->scalarPrototype()
+                    ->validate()
+                        ->always(function ($values) {
+                            if (isset($values['_default'])) {
+                                throw new InvalidDefinitionException("Key '_default' can't be used as build name.");
+                            }
+
+                            return $values;
+                        })
+                    ->end()
                 ->end()
             ->end()
         ;
--- a/src/DependencyInjection/WebpackEncoreExtension.php
+++ b/src/DependencyInjection/WebpackEncoreExtension.php
@@ -13,6 +13,10 @@ use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
+use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\WebpackEncoreBundle\Asset\EntrypointLookup;
 
 final class WebpackEncoreExtension extends Extension
 {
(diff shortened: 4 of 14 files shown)


src/Asset/TagRenderer.php
```diff
@@ -15,18 +15,27 @@
...
+    private $entrypointLookupCollection;
+
     private $packages;
 
-    public function __construct(EntrypointLookupInterface $entrypointLookup, Packages $packages)
-    {
+    public function __construct(
+        EntrypointLookupInterface $entrypointLookup,
+        EntrypointLookupCollection $entrypointLookupCollection = null,
+        Packages $packages
```
weaverryan, line 23: This reordering of the args breaks BC. What if we did this:

1) Replace the first arg entirely with `EntrypointLookupCollection`
2) But, don't add the type-hint for `EntrypointLookupCollection` (as that would be a BC break). Instead, we would do an `instanceof` check in the constructor: if the user is passing an `EntrypointLookupInterface`, we would trigger a BC warning (example: https://github.com/symfony/symfony/blob/63d34515a51819a6a6b3a94d598e8e37068e85ce/src/Symfony/Component/HttpKernel/DataCollector/ConfigDataCollector.php#L36). In this situation, we would instantiate a `new EntrypountLookupCollection()` and pass in just the one `EntrypointLookupInterface` object. If they pass something that is neither of these 2 types, we would throw a `TypeError`

src/DependencyInjection/WebpackEncoreExtension.php
```diff
@@ -26,5 +26,10 @@ public function load(array $configs, ContainerBuilder $container)
 
         $container->getDefinition('webpack_encore.entrypoint_lookup')
             ->replaceArgument(0, $config['output_path'].'/entrypoints.json');
+
+        $container->getDefinition('webpack_encore.entrypoint_lookup_collection')
+            ->replaceArgument(0, array_map(function ($item) {
+                return $item.'/entrypoints.json';
+            }, $config['builds'] ?? []));
```
weaverryan: Let's pass in the "default" path here as well - we could give it a key like `default` (and add some validation in the `Configuration` class to make sure there is no key called `default` set under `builds`

Also, is it possible for `$config['builds']` to not be set? Shouldn't it always be set thanks to the Configuration class (and so the `$config['builds'] ?? []` is unnecessary?)

src/Asset/EntrypointLookupCollection.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        $this->buildEntryPoints = array_map(function ($path) {
+            return new EntrypointLookup($path);
+        }, $buildEntryPoints);
+    }
+
+    public function getEntrypointLookup(string $key): EntrypointLookupInterface
+    {
+        if (!isset($this->buildEntryPoints[$key])) {
+            throw new UndefinedEntrypointException(sprintf('Given entry point "%s" is not configured', $key));
```
weaverryan: I think you're mixing up "entrypoint" and "builds" in a few places. This should be something more like `UndefinedBuildException`, right? And, in `TagRenderer`, the argument to `private function getEntrypointLookup(string $entrypointName = null)` should really be `$buildName`, right?

src/Asset/TagRenderer.php
```diff
@@ -15,18 +15,27 @@
...
+        $this->entrypointLookupCollection = $entrypointLookupCollection ?? new EntrypointLookupCollection([]);
         $this->packages = $packages;
     }
 
-    public function renderWebpackScriptTags(string $entryName, string $packageName = null): string
-    {
+    public function renderWebpackScriptTags(
+        string $entryName,
+        string $entrypointName = null,
+        string $packageName = null
```
weaverryan: Unfortunately, args 2 and 3 need to be reversed, otherwise it's a BC break.

And put the method back on one line (I sometimes do this in my code, but we don't do this in Symfony)

src/Twig/EntryFilesTwigExtension.php
```diff
@@ -34,33 +34,43 @@ public function getFunctions()
...
             ->getCssFiles($entryName);
     }
 
-    public function renderWebpackScriptTags(string $entryName, string $packageName = null): string
-    {
+    public function renderWebpackScriptTags(
+        string $entryName,
+        string $entrypointName = null,
+        string $packageName = null
+    ): string {
```
weaverryan: Same as above: the 2nd and 3rd args need to be rearranged, unfortunately. I don't think there's a good way to "reorder" these args without breaking BC, except adding an entirely new Twig function.

src/Twig/EntryFilesTwigExtension.php
```diff
@@ -34,33 +34,43 @@ public function getFunctions()
...
         return $this->getTagRenderer()
-            ->renderWebpackLinkTags($entryName, $packageName);
+            ->renderWebpackLinkTags($entryName, $entrypointName, $packageName);
     }
 
-    private function getEntrypointLookup(): EntrypointLookupInterface
+    private function getEntrypointLookup(string $entrypointName = null): EntrypointLookupInterface
```
weaverryan: `?string $entrypointName`

Let's allow the value to be null, but force it to be passed to this function. Same in `TagRenderer`

src/Twig/EntryFilesTwigExtension.php
```diff
@@ -34,33 +34,43 @@ public function getFunctions()
...
     {
-        return $this->container->get('webpack_encore.entrypoint_lookup');
+        if (null === $entrypointName) {
+            return $this->container->get('webpack_encore.entrypoint_lookup');
+        }
+        return $this->container->get('webpack_encore.entrypoint_lookup_collection')
+            ->getEntrypointLookup($entrypointName);
```
weaverryan, line 64: Once we make the `EntrypointCollection` also able to handle the "default" build, this should simplify

src/Asset/EntrypointLookupCollection.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+class EntrypointLookupCollection
+{
+    private $buildEntryPoints;
+
+    public function __construct(array $buildEntryPoints)
+    {
+        $this->buildEntryPoints = array_map(function ($path) {
+            return new EntrypointLookup($path);
+        }, $buildEntryPoints);
```
weaverryan: Hmm. Let's tweak this slightly:

1) Make this class receive a `ContainerInterface`, which will be a service locator (like we use in the Twig extension) containing all of the `EntrypointLookup` objects.

2) In `WebpackEncoreExtension`, for each "build", register a new service dynamically. Then, create a container/service locator for all of the builds, and pass that here.

That will allow all `EntrypointLookup` objects to be managed by the container, but also passed lazily to this class, so that we're not instantiating ALL of the `EntrypointLookup` objects just to use one of them.

weaverryan (changes requested):
Hey @shulard!

Wow! Great job with this big feature! I've left a series of comments - it's all just architecture related - avoiding BC breaks, making things lazy for performance, etc. Let's get this guy finished - I think a lot of people are excited for it :).

Cheers!

weaverryan:
Ping @shulard! Are you able to make the changes I suggested?

shulard:
Hello @weaverryan, sorry about the delay but the last week was really busy. I'm able to perform requested changes I hope this will be done today.

src/Asset/EntrypointLookupCollection.php
shulard: Yes you're right, I update.

src/Asset/TagRenderer.php
shulard: Ok, I understand the point about BC breaks :smile:

Only one question about the exception message. For the TypeError, I think:

```
The "$entrypointLookupCollection" argument must be an instance of EntrypointLookupCollection.
```

For the BC warning : 

```
The "$entrypointLookupCollection" argument in method "__construct()" must be an instance of EntrypointLookupCollection.
```

We do you think ?

src/Twig/EntryFilesTwigExtension.php
shulard: Yes, I've removed the lookup / collection usage to rely only on collection.

src/Asset/EntrypointLookupCollection.php
shulard: I've just pushed the changes.

src/Twig/EntryFilesTwigExtension.php
shulard: Since entrypoint name is now always a string (default value is `_default`), I think allowing null is not useful anymore.

src/Twig/EntryFilesTwigExtension.php
shulard: I've reordered the arguments.

src/Asset/TagRenderer.php
shulard: I've reordered the arguments.

src/DependencyInjection/WebpackEncoreExtension.php
shulard: I've updated that part to check for default build name and I removed the `??` operator.

shulard:
Hello @weaverryan I've updated all the requested changes (I think :smile:). 

For me the latest point is the exception messages that maybe need to be refined.

frederickhamla:
Hi !

@weaverryan : Thanks for your work, happy for this feature. Did you know when you merge this request ? 

@shulard : nice job !

src/DependencyInjection/Configuration.php
```diff
@@ -27,6 +27,19 @@ public function getConfigTreeBuilder()
                     ->isRequired()
                     ->info('The path where Encore is building the assets - i.e. Encore.setOutputPath()')
                 ->end()
+                ->arrayNode('builds')
+                    ->useAttributeAsKey('name')
+                    ->scalarPrototype()
+                    ->validate()
+                        ->always(function ($values) {
+                            if (isset($values['_default'])) {
+                                throw new \Symfony\Component\Config\Definition\Exception\InvalidDefinitionException("Key '_default' can't be used as build name.");
```
weaverryan: Minor - add a `use` statement for this

tests/IntegrationTest.php
```diff
@@ -34,6 +48,20 @@ public function testTwigIntegration()
...
+        // styles3.css is not repeated
+        $this->assertNotContains(
+            '<link rel="stylesheet" href="/build/styles3.css">',
+            $html2
+        );
+        // styles4.css is not repeated
+        $this->assertNotContains(
+            '<link rel="stylesheet" href="/build/styles4.css">',
+            $html2
+        );
```
weaverryan, line 78: Hmm, could we split this into a separate test instead of adding to this test? It makes it a bit harder to understand.

tests/IntegrationTest.php
```diff
@@ -75,6 +103,9 @@ public function registerContainerConfiguration(LoaderInterface $loader)
 
             $container->loadFromExtension('webpack_encore', [
                 'output_path' => __DIR__.'/fixtures/build',
+                'builds' => [
+                    'other' =>  __DIR__.'/fixtures/other-build'
```
weaverryan: Instead of `other` - how about `different_build`. The problem is that there is an `other_entry` in the main build, so it's confusing.

weaverryan (changes requested):
This is awesome! Sorry for the delay - only minor comments!

src/DependencyInjection/Configuration.php
shulard: Ok, updated.

tests/IntegrationTest.php
shulard: I've created two test cases: one with contains check and one with "not contains check".

tests/IntegrationTest.php
shulard: Ok, I've renamed it.

weaverryan:
Thank you for your great work Stéphane!

seyfer:
Great PR. :+1: 
Docs need to be updated too and show how multiple entry points can be used.

Also would be great to have v1.0.1 or v1.1.0 release tag with this feature.

seyfer:
@weaverryan Does it break `dev-server --hot`?
See my comment here https://github.com/symfony/webpack-encore/issues/478#issuecomment-453978167

nyroDev:
I know this have been merged but not yet published, but I found an inconsistency between entrypoints and manifest configuration.

In assets, the configuration might looks like:
```
framework:
    assets:
        json_manifest_path: '%kernel.project_dir%/public/build/manifest.json'
        packages:
            mobile:
                json_manifest_path: "%kernel.project_dir%/public/buildMobile/manifest.json"
```

And in webpack_encore, with this PR, this might look like :  
```
webpack_encore:
    # The path where Encore is building the assets.
    # This should match Encore.setOutputPath() in webpack.config.js.
    output_path: '%kernel.project_dir%/public/build'
    builds:
        mobile: '%kernel.project_dir%/public/buildMobile'
```

My first suggestion would be to simply rename `builds` into `packages` directly. (I could PR it if needed)

And then, maybe we can have a more global thinking later in order to merge both configuration and to have only 1 place to configure the output folder.
One solution might be that webpack_encore bundle read the json_manifest_path and deduce the entrypoints path from it.

shulard:
@nyroDev, for me this is two different things. Your first code sample is related to the `Asset` component configuration. `package` mean asset package here so you can define multiple asset stores to be used in your application.

Your second sample is the `webpack_encore` bundle configuration. In that context, the `build` term mean a specific webpack build.

For me those two terms are relevant in their specific context. Event with that PR merged, you can use different asset packages with different webpack builds. Each of these features allow to control specific part of the asset management in your app.

ureimers:
Hi @shulard,
you're using the parameter `%kernel.public_dir%` but that one's not defined in Symfony. At least not in Symfony 4.2:
`You have requested a non-existent parameter "kernel.public_dir".` It's not hard, to change that to '%kernel.project_dir%/public' but I'm just curious.

Am I missing something?

shulard:
Hi @ureimers,

When are you getting that error ? I don't find any usage of `public_dir` in the code, only in the README which maybe contains a bad example...

ureimers:
@shulard, no you're right, I just took it from the README (and your initial post from this PR) and thought that maybe my SF 4.2 was incorrectly setup. It was just out of curiosity.

Cheers

shulard:
Ok :smile:

It's an important feedback if the variable doesn't exists, I'll make a tiny PR to update the README.

shulard:
@ureimers, I've created a tiny PR to update the README and I've updated that PR description to avoid confusion :wink:. Thank you for the feedback !

ureimers:
@shulard, you're welcome and thank you for the quick response!

PR #18, opened 2018-11-16
Remove trailing slash on link tag

darrylhein:
Though allowed, generally in HTML5 it's written without it. Based on the style of <script> tag, I'm guessing the bundle isn't supporting anything but HTML5.


weaverryan:
Thanks! But duplicate of #15... but I will for sure merge it :).

Cheers!

PR #21, opened 2018-12-05
Catch error with entrypointlookups in sub-requests

### LINKED ISSUE #910: Tricky error related to 500 error pages
Yesterday I merged #909 as an (ugly) temporary solution to a very tricky error.

How to reproduce the error:

* Use `prod` and `debug=0` in `.env`
* Comment the `DATABASE_URL` config to trigger a 500 error
* The 500 error page sometimes is styled with CSS and sometimes isn't:
  * The 500 error for this page shows styles: `/en/blog/posts/lorem-ipsum-dolor-sit-amet-consectetur-adipiscing-elit`
  * The 500 error for this page doesn't: `/en/blog/`

Can anyone spot the reason and provide a pull request to fix it? Thanks!

ckrack:
In EntrypointLookup, the list of previously returned entries is not reset
on sub-requests.
This leads to errors when sub-requests return full html pages.

Fixes symfony/demo#910

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the Symfony WebpackEncoreBundle package.
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\WebpackEncoreBundle\EventListener;
+
+use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
+use Symfony\WebpackEncoreBundle\Asset\EntrypointLookup;
+use Twig_Error_Runtime;
+
+class ExceptionListener
+{
+    private $entrypointLookup;
+
+    public function __construct(EntrypointLookup $entrypointLookup)
+    {
+        $this->entrypointLookup = $entrypointLookup;
+    }
+
+    public function onKernelException(GetResponseForExceptionEvent $event)
+    {
+        $exception = $event->getException();
+
+        // Reset the entrypointLookup list of previously returned entries, as Twig_Error_Runtime will initialise a sub-request
+        if ($exception instanceof Twig_Error_Runtime) {
+            $this->entrypointLookup->reset();
+        }
+    }
+}
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -9,10 +9,11 @@
 
 namespace Symfony\WebpackEncoreBundle\EventListener;
 
-use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
 use Symfony\WebpackEncoreBundle\Asset\EntrypointLookup;
+use Twig_Error_Runtime;
 
-class RequestListener
+class ExceptionListener
 {
     private $entrypointLookup;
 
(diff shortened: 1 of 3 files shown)


Lyrkan:
I'm wondering if that's the right way to handle that... if a sub-request is created but then popped off the stack you'd go back to the original one with the wrong returned entries, right?

Wouldn't it be better to keep track of the returned entries per request instead?

ckrack:
> I'm wondering if that's the right way to handle that... if a sub-request is created but then popped off the stack you'd go back to the original one with the wrong returned entries, right?

Yes, I think so.
It will also have another issue: when a sub-request returns only partial html, but wants to output an additional asset in there, due to the reset all assets would be output again.

> Wouldn't it be better to keep track of the returned entries per request instead?
Not sure how this would solve it. When a sub-request overrides the output, it should include all necessary assets. When it just adds some html, it could include more assets but should not include the ones previously output.

Maybe the more specific fix from 6b6b731 is better, as it only covers the use case with `Twig_Error_Runtime`.

weaverryan:
Yea, this is a complex issue - thanks for checking into it @ckrack.

> Wouldn't it be better to keep track of the returned entries per request instead?

Yes, I *also* think this is the best way to do this. I was talking with @ckrack about this, and I thought for a moment that this was NOT the best way - but now I think it is. Basically. the EntrypointLookup should keep track of which assets have been used on a request-by-request basis. We can probably do this by injecting the RequestStack and using the `spl_object_hash()` as a key for tracking which assets have been returned for that specific request.

There is still *another* edge-case, but I think it can just be handled by the user (and some documentation): if you render 2 full HTML pages within *one* sub-request (e.g. render HTML for a PDF, then render the normal template - odd case, but possible), you will need to manually reset the stack.

@ckrack are you able to make the above changes? Thanks

weaverryan:
Ping @ckrack! I just merged another PR, which would have probably conflicted with the work that needs to be done here. So, things are clear now. Do you have some time to work on this?

Thanks for the attention!

weaverryan:
Ok, on *third* thought, I think this is the right approach - we cannot predict what behavior is correct for sub-requests in general.

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        $this->entrypointLookup = $entrypointLookup;
+    }
+
+    public function onKernelException(GetResponseForExceptionEvent $event)
+    {
+        $exception = $event->getException();
+
+        // Reset the entrypointLookup list of previously returned entries, as Twig_Error_Runtime will initialise a sub-request
+        if ($exception instanceof Twig_Error_Runtime) {
```
weaverryan, line 30: Why this error specifically? All exceptions will result in a full page being rendered (by default). So, I *think* what you’re doing is trying to detect if the error came from Twig, which is the most likely cause of a situation where Encore assets have been rendered already... and THEN an exception happens. Am I right?

If so... I’m not sure it’s a good idea. Regardless of the error, if we try to render a template for an error page, that should probably use a fresh set of Encore assets.

weaverryan: I’m also trying to think if we should limit this to the master request... does the default exception behavior (rendering an error template) happen only on the master request? That seems likely - I just haven’t checked the code yet.

weaverryan: I've thought more about this. I think we should:

1) Always reset on exception - not just if there was a Twig error. There is going to be some edge cases (people successfully rendering in Twig, then a listener throws an exception after).

2) We should reset on all requests, not just the master request.

There may still be edge cases - but let's do that and see what happens. Fortunately, this is in "exception" land - so it's already not a normal request.

weaverryan:
Also, there is now one more complication: #17 added support for multiple builds. So there is now an `EntrypointLookupCollection` that can contain multiple `EntrypointLookup`, and we need to clear all of them (probably will need to add a method to be able to get all of them from that new class).

weaverryan:
Ping @ckrack! Are you still able to work on this? Thanks

emodric:
Hi @weaverryan ! What can be done to move this forward? Can we help?

weaverryan:
Closing in favor of #74 - thanks for getting us moving on this @ckrack!

PR #24, opened 2018-12-24
Tweak docs: Now .splitEntryChunks() is enabled by default

bocharsky-bw:
Now it's already enabled in the recipe, see https://github.com/symfony/recipes/pull/509


weaverryan:
Nice one Victor!

PR #27, opened 2019-01-12
Extend from the ResetInterface

fionera:
This way the Container automaticly resets the returnedFile array and this hack is obsolete https://github.com/php-pm/php-pm-httpkernel/pull/151

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         "symfony/asset": "^3.4 || ^4.0",
         "symfony/config": "^3.4 || ^4.0",
         "symfony/dependency-injection": "^3.4 || ^4.0",
-        "symfony/http-kernel": "^3.4 || ^4.0"
+        "symfony/http-kernel": "^3.4 || ^4.0",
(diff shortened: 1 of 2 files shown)


weaverryan:
This is definitely a good idea. But, it would mean that the bundle should now require symfony/contracts at version 4.2 I believe.

fionera:
It already does... http-kernel requires event-dispatcher wich depends on contracts

weaverryan:
Ah. But now we have a direct dependency on it - we shouldn’t rely on it indirectly. Also, we will need to specifically require version 4.2 or higher, not just 4.0 or higher (currently, if we don’t explicitly require it in this lib, someone could get 4.1 of everything, which does not rely on contracts).

This will also make this bundle work for Symfony 4.2 only (well, at least 4.2 of the contracts library will be required).

nicolas-grekas:
Contracts is a standalone package so the bundle will still be compatible with 4.0

weaverryan:
Thank you @fionera!

PR #37, opened 2019-01-28
Allow disabling default build if multiple entry points are defined

### LINKED ISSUE #32: Optional default entrypoint when using multiple entrypoints
I'm trying out multiple entrypoints, and first test was to define a single `builds` directive:

```
webpack_encore:
    builds:
        frontend: '%kernel.project_dir%/web/build'
```

However, Webpack Encore complains that the `output_path` is not configured: `The child node "output_path" at path "webpack_encore" must be configured.`. Is this the intended behaviour? How would I go about specifying this kind of config as a starting point for other developers on my team to extend with more entrypoints?

emodric:
Fixes #32.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ file:
 webpack_encore:
     # The path where Encore is building the assets - i.e. Encore.setOutputPath()
     output_path: '%kernel.project_dir%/public/build'
+    # If multiple builds are defined (as shown below), you can disable the default build:
+    # output_path: false
 
     # if you have multiple builds:
     # builds:
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -23,6 +23,12 @@ final class Configuration implements ConfigurationInterface
         $rootNode = method_exists($treeBuilder, 'getRootNode') ? $treeBuilder->getRootNode() : $treeBuilder->root('webpack_encore');
 
         $rootNode
+            ->validate()
+                ->ifTrue(function (array $v): bool {
+                    return false === $v['output_path'] && empty($v['builds']);
+                })
+                ->thenInvalid('Default build can only be disabled if multiple entry points are defined.')
+            ->end()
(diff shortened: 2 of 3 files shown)


weaverryan (changes requested):
Hey @emodric!

Sorry for the slow attention on this. I merged another PR and created conflicts with this one - could you check those?

I'm +1 for this PR. However, I'm interested in what the error is if you disable the default entrypoint, and then accidentally use a feature without specifying which build you want - I think the user will see errors like "The build _default does not exist" - which isn't a great error.

weaverryan:
#45 will also help and affect this PR

emodric:
@weaverryan Will get back to you next week, can't get to it any sooner.

emodric:
@weaverryan I've pushed the rebased commit. As for #45, how will it help and affect this PR?

When someone uses a default build which has been disabled, the error message is: `Given entry point "_default" is not configured`.

Maybe we could try to determine which builds are configured and offer them as an alternative in the error message, e.g: `The default build is not enabled. Use one of "frontend", "backend", "app1", "app2" builds.`

jRexhmati:
Hello @emodric do you plan to merge this pr anytime soon?

emodric:
You should probably be asking @weaverryan that :)

jRexhmati:
@emodric  Thanks for the info :)
@weaverryan Do you plan to merge this soon? I can give you a hand ✋if you need

weaverryan (approved):

weaverryan:
Sorry about the slow merge on this - it's a really great PR. Even the README got updated! Thank you Edi!

emodric:
@weaverryan Thanks for the merge :)

However, what about this part, that you were initially concerned about?

> Maybe we could try to determine which builds are configured and offer them as an alternative in the error message, e.g: The default build is not enabled. Use one of "frontend", "backend", "app1", "app2" builds.

PR #38, opened 2019-01-29
Remove unknown variable `%kernel.public_dir%` from README.md.

shulard:
As said in https://github.com/symfony/webpack-encore-bundle/pull/17#issuecomment-456441357, this variable does not exists in Symfony 4.2 and must not be used in the README as a valid example.


weaverryan:
Thanks Stéphane!

PR #39, opened 2019-02-01
Fix webpack_encore.entrypoint_lookup registration

### LINKED ISSUE #36: Exception thrown when retrieving `webpack_encore.entrypoint_lookup` from container
### Problem
Prior to v1.1.0 it was possible to fetch `webpack_encore.entrypoint_lookup` from the container to interact with the `EntrypointLookup` class.  Now when fetching `webpack_encore.entrypoint_lookup` from the container we get the following exception:

```
FatalExceptionError
Type error: Argument 1 passed to Symfony\WebpackEncoreBundle\Asset\EntrypointLookup::__construct() must be of the type string, object given, called in /vagrant/var/cache/prod/Container3svdnsj/getWebpackEncore_EntrypointLookupService.php on line 8
```

Looking at `WebpackEncoreExtension` the following line appears to be the culprit:

```php
$container->getDefinition('webpack_encore.entrypoint_lookup')
    ->replaceArgument(0, $factories['_default']);
```

Not being very familiar with the container and the way it works I can only speculate that it's not being resolved from the container correctly and passing the object itself rather than the output path.

Calling `webpack_encore.entrypoint_lookup[_default]` works fine as a temporary workaround but it's not an ideal solution.

### Info
**PHP:** 7.1.8
`symfony/symfony`: 3.4.20
`symfony/webpack-encore-bundle`: v1.1.0

MrMitch:
The `webpack_encore.entrypoint_lookup` service instanciation was throwing an exception, preventing it from being used properly with dependency injection.

Closes #36

### DIFF
--- a/src/DependencyInjection/WebpackEncoreExtension.php
+++ b/src/DependencyInjection/WebpackEncoreExtension.php
@@ -36,7 +36,7 @@ final class WebpackEncoreExtension extends Extension
         };
 
         $container->getDefinition('webpack_encore.entrypoint_lookup')
-            ->replaceArgument(0, $factories['_default']);
+            ->replaceArgument(0, $config['output_path'].'/entrypoints.json');
         $container->getDefinition('webpack_encore.entrypoint_lookup_collection')
             ->replaceArgument(0, ServiceLocatorTagPass::register($container, $factories));
     }
--- a/tests/DependencyInjection/DummyService.php
+++ b/tests/DependencyInjection/DummyService.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Symfony\WebpackEncoreBundle\Tests\DependencyInjection;
+
+
+use Symfony\WebpackEncoreBundle\Asset\EntrypointLookup;
+
+class DummyService
+{
+    protected $entryPointLookup;
+
+    public function __construct(EntrypointLookup $entrypointLookup)
+    {
+        $this->entryPointLookup = $entrypointLookup;
+    }
+
+    public function getEntryPointLookup(): EntrypointLookup
+    {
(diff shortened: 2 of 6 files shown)


MrMitch:
The tests pass, the failure is due to a deprecation warning that was not introduced by this PR :

```
The "PHPUnit_Framework_MockObject_Matcher_ConsecutiveParameters::verify()" method will
require a new "PHPUnit_Framework_MockObject_Invocation $invocation" argument in the next
major version of its parent class "PHPUnit_Framework_MockObject_Matcher_StatelessInvocation", 
not defining it is deprecated.
```

weaverryan:
Hey @MrMitch!

I noticed this a few days ago too - that service is totally broken. The reason we didn't notice is that... if you look closely, it's not used inside the bundle anymore! It's even passed to `webpack_encore.twig_entry_files_extension`, but that service doesn't actually use it.

The proper solution is to not use this service at all anymore - and to use `webpack_encore.entrypoint_lookup_collection` instead and then fetch out the default entrypoint lookup.

Would you mind re-working this PR to actually completely remove the `webpack_encore.entrypoint_lookup` service?

weaverryan:
Thanks for opening this @MrMitch. As I mentioned earlier, we've actually removed the broken service entirely. This was a really nice PR though - I like the integration tests you added. I hope we'll see you again here :)

MrMitch:
Hi @weaverryan, thank you for the feedback. I've had no time to circle back to this and the solution you implemented in #45 seems more elegant and more flexible.
Any reason why it couldn't get included in the recent 1.2.0 or 1.2.1 releases ?

Issue #40, opened 2019-02-13
Unable to use dashes in build name due to array key normalization


smoelker:
The build name in the example below will change from ```some-name``` to ```some_name``` due to array key [normalization](https://symfony.com/doc/current/components/config/definition.html#normalization) of the configuration component.

```yaml
webpack_encore:
    output_path: '%kernel.project_dir%/public/static/'
    builds:
        some-name: '%kernel.project_dir%/public/static/some-name'
```

Resulting in an UndefinedBuildException: 

```An exception has been thrown during the rendering of a template ("Given entry point "some-name" is not configured").```

When I call 
```twig
  {{ encore_entry_link_tags('frontend', null, 'some-name' }}
```

It does work when I use the following line:
```twig
  {{ encore_entry_link_tags('frontend', null, 'some_name' }}
```

I don't see any reason to normalize the keys of the builds config. I should be free to choose whatever name I want for my build, right?

PR #41, opened 2019-02-13
Use normalizeKeys(false) for builds config

smoelker:
Fix for #40 to allow dashes in the build names.


weaverryan:
I agree! No reason to normalize the build names - thanks @smoelker!

PR #44, opened 2019-02-26
fixed a little issue

d4wae89d498:
This issue was causing my javascript and css tags to stop rendering after 2/3 pages refresh on Swoole.
I hope that I will help someone. Also if I missed something please tell it to me.

Thank you for your tool, it is very useful.


weaverryan:
Hey @MarcFaussurier!

I'm not familiar with Swoole, but my guess is that you only boot the kernel once, and then serve many requests. There's not a bug in the library, but in that situation, you need to reset() the entrypoints cache after service each request. 

See #27 for how it was solved in php-pm.

Cheers!

PR #45, opened 2019-03-01
Adding Autowireable interfaces & docs for reset()

### LINKED ISSUE #33: encore_entry_css_files returns nothing if called multiple time
I'm using this function to render PDF with inline CSS as it was suggested in https://github.com/symfony/webpack-encore-bundle/issues/16#issuecomment-455024938:

```twig
<style>{{ asset_get_content(encore_entry_css_files('pdf')) }}</style>
```

But if I call the twig `render` method multiple time (to render multiple PDFs on one command), only the first will get CSS.

For the other, the css file list is just empty.

So I tried this for test on my homepage:

```twig
{{ dump(encore_entry_css_files('pdf')) }}
{{ dump(encore_entry_css_files('pdf')) }}
{{ dump(encore_entry_css_files('pdf')) }}
```

And it confirms the issue:

image

To me, it's a bug. Why this function would not return the needed files if call multiple times?

weaverryan:
Fixes #33 by making it reasonable to reset() your assets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,3 +92,39 @@ and `build/entry1.js`, then `encore_entry_script_tags()` is equivalent to:
 If you want more control, you can use the `encore_entry_js_files()` and
 `encore_entry_css_files()` methods to get the list of files needed, then
 loop and create the `script` and `link` tags manually.
+
+## Rendering Multiple Times in a Request (e.g. to Generate a PDF)
+
+When you render your script or link tags, the bundle is smart enough
+not to repeat the same JavaScript or CSS file within the same request.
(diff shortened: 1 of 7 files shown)


Nyholm (approved):
I’m not 100% familiar with the source of this bundle. But I think this PR is okey. 

Btw, is there a reason why you made one PR with two unrelated features?

MrMitch:
I really love this, it would make autowiring and autoconfiguring way easier for those of us who need to `reset` the EntryPointLookup (when used multiple times in the same request).

@weaverryan do you need any help on this to get it released as soon as possible ? I'd be glad to help.

PR #50, opened 2019-03-04
integration test created for cache warmer

codayblue:
- Created integration test
- Created an alias so that cash warmer could be autowired
- Fixed a bug where entrypoints.json was not added to the path so the exception would get thrown that entry point was missing


weaverryan:
Thanks Cody!

PR #53, opened 2019-03-15
Improved entrypointlookupcollection

quisse:
Made entrypointlookup argument optional and use `_default` entrypoint by default.


weaverryan:
Thanks for this! But I think #45 will accomplish what you need already :)

Issue #43, opened 2019-02-22
Don't require assets in test environment


karser:
Hello,

Twig functions encore_entry_link_tags / encore_entry_script_tags throw an exception if assets are not compiled:
```
An exception has been thrown during the rendering of a template ("Could not find the entrypoints file from Webpack: the file "/var/www/app/public/build/entrypoints.json" does not exist.")
```
This behavior is annoying when it comes to functional testing. Because building assets for symfony crawler tests is a waste of time.
Currently, I'm using this workaround
```
    <head>
        {% if app.environment != 'test' %}{{ encore_entry_link_tags('app') }}{% endif %}
    </head>
    <body>
        {% if app.environment != 'test' %}{{ encore_entry_script_tags('app') }}{% endif %}
    </body>
```

It would be nice to have an option not to throw exceptions for the test environment.

weaverryan:
Sure, this makes sense. We could add a `strict_mode` option or something that defaults to `true`, but could be set to false to simply not render assets if they are not there.

If you're interested, you can open a PR.

Cheers!

karser:
Thanks for the feedback. I'll do that when I have free time

axzx:
@karser 

I am looking for resolve exception
`Asset manifest file "public/build/manifest.json" does not exist.`
when I run test and I dont have directory build

Do you know how run test on github actions without yarn build?

karser:
@axzx I'm using this script
`bin/run-tests.sh`
```
#!/usr/bin/env bash

set -e

if [[ ! -e ./public/build/manifest.json ]]; then
    mkdir -p ./public/build
    echo "{}" > ./public/build/manifest.json
fi

composer install --dev --prefer-dist --no-interaction --no-scripts --no-progress --no-suggest
vendor/bin/simple-phpunit -c ./phpunit.xml.dist
```

daFish:
You can also define a custom `json_manifest_path` for the test-environment:
```yaml
# config/test/framework.yaml
framework:
    test: true
    session:
        storage_id: session.storage.mock_file
    assets:
        json_manifest_path: ~
```

JohJohan:
When i have the config mentioned by @daFish i get the following error when i dont have assets installed:
```
[CRITICAL] Uncaught PHP Exception Twig\Error\RuntimeError: "An exception has been thrown during the rendering of a template ("Could not find the entrypoints file from Webpack: the file "/x/public/build/entrypoints.json" does not exist.")."
```

[...]

PR #54, opened 2019-03-20
Added strict mode #43

karser:
This PR introduces `strict_mode` which is true by default so it doesn't change the original bundle behavior. If set to false it suppresses all the exceptions thrown because of missing builds. It makes sense to disable strict mode for test environment where you don't need to build your assets.

```
# config/packages/test/webpack_encore.yaml
webpack_encore:
    strict_mode: false
```
Here is the original issue https://github.com/symfony/webpack-encore-bundle/issues/43


karser:
Hi @weaverryan , Do you think it can be merged? Please take a look at this PR

src/Resources/config/services.xml
```diff
@@ -27,6 +27,7 @@
                     </argument>
                 </service>
             </argument>
+            <argument />
```
weaverryan: Add a comment here: `<!-- strict mode -->` (but also, I think we may move this config to another service)

src/Twig/EntryFilesTwigExtension.php
```diff
@@ -18,10 +18,12 @@
 final class EntryFilesTwigExtension extends AbstractExtension
 {
     private $container;
+    private $strictMode;
 
-    public function __construct(ContainerInterface $container)
+    public function __construct(ContainerInterface $container, bool $strictMode)
```
weaverryan: I think this needs to be an argument on `EntrypointLookup` - that's where the exception originates - and strict mode should also apply if someone were using the service directly for some reason.

weaverryan (changes requested):
A few implementation things, but this is definitely a good idea - don't see any reason not to do this :)

src/Twig/EntryFilesTwigExtension.php
karser: Yeah, I tried to do that at first iteration. I required $strictMode in EntrypointLookup and it turned out into patching almost every method with the strict mode logic and still it continued to throw exception in `EntrypointLookupCollection::getEntrypointLookup` because it must return `EntrypointLookupInterface` but there was no result there. I started to think whether I should change its signature to allow null or return a EntrypointLookupInterface stub there. Then I gave up that and ended up with a more elegant solution - just catch exceptions in 5 twig functions and that's it.
Here is what the original solution looked like (left side):
image

weaverryan: Hmm. I understand. But I still think we need to do it the uglier way, because if we merge it how it is now, someone will eventually file a bug that strict mode only works in Twig, and not directly with the services (and we do support using the services directly).

Your patch in `EntrypointLookup` makes sense to me - if the file doesn't exist, just pretend like it's a file full nothing - no entries.

> still it continued to throw exception in EntrypointLookupCollection::getEntrypointLookup because it must return EntrypointLookupInterface but there was no result there

I don't quite understand this part. Even if the `entrypoints.json` file doesn't exist, the `EntrypointLookupCollection` should still be configured to know which "build names" are valid and should still be populated with those `EntrypointLookup` objects. Unless your code is trying to find a build called `foo_bar`, and that's not configured under the `builds` section for this bundle. But that *should* be invalid. I think I'm missing something here... Let me know :)

karser: I see your point - `EntrypointLookupCollection::getEntrypointLookup` fails because the build is not configured but not because the build file doesn't exist. That is, I modified test `$kernel->builds = ['different_build' => 'missing_build'];` so now it knows about `different_build` and tests pass.
I moved the entire logic to the entrypoint lookup classes instead of twig extension.

karser: @weaverryan makes sense?

src/Asset/EntrypointLookup.php
```diff
@@ -29,11 +29,15 @@ class EntrypointLookup implements EntrypointLookupInterface, IntegrityDataProvid
 
     private $cache;
 
-    public function __construct(string $entrypointJsonPath, CacheItemPoolInterface $cache = null, string $cacheKey = null)
+    private $strictMode;
+
+    public function __construct(string $entrypointJsonPath, CacheItemPoolInterface $cache = null,
+                                string $cacheKey = null, bool $strictMode = true)
```
weaverryan: Single line here

src/Asset/EntrypointLookup.php
```diff
@@ -113,6 +117,7 @@ private function getEntriesData(): array
         }
 
         if (!file_exists($this->entrypointJsonPath)) {
+            if (!$this->strictMode) return [];
```
weaverryan:
```
if (!$this->strictMode) {
    return [];
}

throw new \InvalidA...
```

src/DependencyInjection/Configuration.php
```diff
@@ -38,6 +38,10 @@ public function getConfigTreeBuilder()
                     ->info('Enable caching of the entry point file(s)')
                     ->defaultFalse()
                 ->end()
+                ->booleanNode('strict_mode')
+                    ->info('Throw exceptions if the entry point file(s) is missing')
```
weaverryan: > Throw an exception if the `entrypoints.json` file is missing or an entry is missing from the data.

src/DependencyInjection/WebpackEncoreExtension.php
```diff
@@ -50,13 +50,20 @@ public function load(array $configs, ContainerBuilder $container)
 
         $container->getDefinition('webpack_encore.entrypoint_lookup_collection')
             ->replaceArgument(0, ServiceLocatorTagPass::register($container, $factories));
+
         $container->setAlias(EntrypointLookupInterface::class, new Alias($this->getEntrypointServiceId('_default')));
     }
 
-    private function entrypointFactory(ContainerBuilder $container, string $name, string $path, bool $cacheEnabled): Reference
+    private function entrypointFactory(ContainerBuilder $container, string $name, string $path,
+                                       bool $cacheEnabled, bool $strictMode): Reference
```
weaverryan: One line

weaverryan (changes requested):
Looks great - minor comments (and needs a rebase now)

karser:
@weaverryan fixed and rebased.

karser:
@weaverryan fixed and rebased.

karser:
@weaverryan Have you had time to look into this? I think it's ready for merge

weaverryan:
Very well done! Thanks for the great work and your patience @karser!

karser:
I'm glad it's merged! :+1:

Issue #10, opened 2018-10-29
Allow custom script/link tag attributes to be passed


weaverryan:
There are a few legit use-cases for wanting to use the `encore_entry_script_tags()` functions, but pass a custom attribute:

* `integrity` attribute
* `preload`, `async`, etc

Another attribute, in theory, would be adding `data-` attribute on your script tag. But, that doesn't make sense with `encore_entry_script_tags()`, because this could render *multiple* script tags.

I think the main use-cases then are the ones in the list above. And, these might be best integrated (especially `integrity`) with some global option. I'd like to get use-case feedback and use that to motivate how this is designed.

dunglas:
It could be nice to be able to integrate `encore_entry_script_tags()` with the WebLink component too, to allow the use to adk the web server to push the assets when using a HTTP/2 connection.

nicolas-grekas:
💯 with Kévin!

alexismarquis:
Having an option for generating absolute urls would be useful as well. In my case I need these for wkhtmltopdf rendering.

iquito:
A way to just get a list of files would be nice - `encore_entry_script_tags()` only covers one use case, instead having something like `encore_script_files()` in Twig which returns an array of files you can then do whatever you want with, and also a service in PHP to get these files, so one can generate script tags in any way needed and (for instance) do preloading or other HTTP things in a controller.

For Content-Security-Policy 3, nonce and strict-dynamic one would also need custom script tags (to add another use case) - a generic way of letting anybody do what they want would be most flexible, because I don't think there is a finite set of use cases you can completely cover when you only offer to specifically generate the script tags (or, for example, specifically add it to WebLink) via the bundle.

iquito:
Ah I just saw that those already exist - `encore_entry_js_files` and `encore_entry_css_files`. Having a clearly visible tutorial on how to use those to generate custom HTML code and also use the service in a controller (or somewhere else) would solve the problem better then having a more complicated `encore_entry_script_tags()`, in my opinion, but both are of course possible.

Maybe it is also worth mentioning `encore_entry_js_files` and `encore_entry_css_files` in the blog article on symfony.com as an alternative to the new syntax, as it is "closer" to the old way of doing it and more flexible. I saw now that it is mentioned in the comments, but not all people might read all the comments or might overlook it.

stof:
yeah, the 2 functions returning the list of files should be documented, as they allow full flexibility.

GenieTim:
[...]

PR #56, opened 2019-04-01
propose fix for CORS issues with missing crossorigin tag

### LINKED ISSUE #55: CORS issues with new integrity feature.
The latest version 1.4.0 with the Encore. enableIntegrityHashes() function from @symfony/webpack-encore version 0.26.0

Generates code like:

```

<link rel="stylesheet" href="https://cdn.example.com/webpack/css_main_header.03d68e17.css?v=1554057765" integrity="sha384-2ZeLK3oSpOQmeB6N947S85lqfOVaKdknVpsVOJ0u+F6sohBFltWQ/bLpwAcjayoi">

<script src="https://cdn.example.com/webpack/runtime.6601bbb6.js?v=1554057765" integrity="sha384-cEq/N+NkEWk8qaypIggtQLcs8ib4ZmONrlkGkwJadcQ3+oXVSXHLh49cUXCOGjje"></script>

```

which when deployed will result in a broken site and red in the console log of chrome with message like 

```
...has an integrity attribute, but the resource requires the request to be CORS enabled to check the integrity, and it is not. The resource has been blocked because the integrity cannot be enforced.
```

This is because the tags are missing the attribute

```
crossorigin="anonymous"
```

(View the source of this github page to see how github do this) 

This makes enableIntegrityHashes() currently useless in Symfony 4. 

// cc  https://github.com/symfony/webpack-encore-bundle/pull/42  @Lyrkan

PhilETaylor:
This fixes https://github.com/symfony/webpack-encore-bundle/issues/55 and comes with the unit tests too ;-) 

updated:

This PR also allows the value for crossorigin attribute to be defined in the configuration yaml


Lyrkan:
Hi @PhilETaylor,

Indeed, that's an issue in a cross-domain context (which is not always the case).

I'm not against adding `crossorigin="anonymous"` but are we sure that other people with different setups won't want/need to use other (or no) values for that attribute?

I would 100% agree with using `anonymous` as the default value if there was also a way to override it, either by adding something to `enableIntegrityHashes` or directly when calling `{{ encore_entry_script_tags }}` (kinda related: https://github.com/symfony/webpack-encore-bundle/issues/10)

PhilETaylor:
> Indeed, that's an issue in a cross-domain context (which is not always the case).

Its the case 100% of the time if you are using a CDN to serve your assets :) 

I dont have any strong opinion either way, however when I upgraded and deployed to production (which uses a cdn) it broke my site.. hence I had to work out why, document it, and find a solution.. which I have proposed. 

The current situation is that anyone that follow what I did, updates, enables integrity, and then deploys with assets from a cdn will break their site...

PhilETaylor:
there are only two options really for the crossorigin attribute. https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes

`anonymous` and `use-credentials`

I've never seen anyone use the latter... but at the moment this feature is BROKEN because deploying assets on a CDN with the integrity attribute and no crossorigin attribute **WILL BREAK THE PRODUCTION SITE!**  (break == asset loading will be blocked by Google Chrome)

I would say a quick fix would be the hard coding of crossorigin as proposed by this PR and then further discussion on making it configurable can be had later, it seems #10 has been open since last year with no progress and you cant expect to leave this feature broken for the majority while someone makes that decision

PhilETaylor:
ok with 768a820 I have now made the value of crossorigin attribute configurable  :-) 

if none is set in the config, then it still defaults to anonymous

PLEASE TEST!

PhilETaylor:
flex recipe also updated here https://github.com/symfony/recipes/pull/562

src/DependencyInjection/Configuration.php
```diff
@@ -34,6 +34,9 @@ public function getConfigTreeBuilder()
                     ->isRequired()
                     ->info('The path where Encore is building the assets - i.e. Encore.setOutputPath()')
                 ->end()
+                ->scalarNode('crossorigin')
```
stof: as there is only 2 valid values, I suggest using an `enumNode` instead

src/DependencyInjection/WebpackEncoreExtension.php
```diff
@@ -51,6 +51,9 @@ public function load(array $configs, ContainerBuilder $container)
         $container->getDefinition('webpack_encore.entrypoint_lookup_collection')
             ->replaceArgument(0, ServiceLocatorTagPass::register($container, $factories));
         $container->setAlias(EntrypointLookupInterface::class, new Alias($this->getEntrypointServiceId('_default')));
+
+        $container->getDefinition('webpack_encore.tag_renderer')
+            ->replaceArgument(2, $config['crossorigin']??'anonymous');
```
stof: The default value should be moved to the Configuration class. This way, tooling can now about it.

src/Resources/config/services.xml
```diff
@@ -16,6 +16,7 @@
         <service id="webpack_encore.tag_renderer" class="Symfony\WebpackEncoreBundle\Asset\TagRenderer">
             <argument type="service" id="webpack_encore.entrypoint_lookup_collection" />
             <argument type="service" id="assets.packages" />
+            <argument type="string" id="webpack_encore.crossorigin" />
```
stof: `id` is not a valid attribute for `<argument type="string">` (and we generally omit `type="string"` entirely as that's the default behavior too)

stof (changes requested):

PhilETaylor:
@stof Thanks for the comments... coding on the sofa on laptop is not my best place to code :) also still learning some of the deep symfony stuff :) ... "it works" is not enough :) I'll go grab a coffee and update this PR with your comments - thanks for the feedback

src/DependencyInjection/WebpackEncoreExtension.php
PhilETaylor: Done.

src/Resources/config/services.xml
PhilETaylor: I think I have done this right now, just `<attribute>anonymous</attribute>` ???

src/DependencyInjection/Configuration.php
PhilETaylor: Done.

PhilETaylor:
@stof c304692 has changes based on your feedback, only thing Im not sure is if the services xml should be 

`<argument/>`

or the repetition of the default value like 

`<argument>anonymous</argument>`

I could not find any examples to learn from on this :)

PhilETaylor:
ok I have settled on 

`<argument /><!-- crossorigin -->`

as I found that example in symfony/security-bundle like that :)

PhilETaylor:
This is ready for testing/feedback again now :)

src/DependencyInjection/Configuration.php
```diff
@@ -34,6 +34,11 @@ public function getConfigTreeBuilder()
                     ->isRequired()
                     ->info('The path where Encore is building the assets - i.e. Encore.setOutputPath()')
                 ->end()
+                ->enumNode('crossorigin')
+                    ->defaultValue('anonymous')
+                    ->values(['anonymous', 'use-credentials'])
```
Lyrkan: I'm not sure if this would be useful or not but should we also handle a `false` case in which the attribute is not added?

src/Asset/TagRenderer.php
```diff
@@ -18,9 +18,12 @@ final class TagRenderer
 
     private $packages;
 
+    private $crossorigin;
+
     public function __construct(
         $entrypointLookupCollection,
-        Packages $packages
+        Packages $packages,
+        string $crossorigin
```
Lyrkan: Maybe it would be better to have a `$defaultAttributes` array here instead? in case we want to be able to configure more of them later?

README.md
```diff
@@ -24,7 +24,10 @@ webpack_encore:
     output_path: '%kernel.project_dir%/public/build'
     # If multiple builds are defined (as shown below), you can disable the default build:
     # output_path: false
-
+    
+    # if using Encore.enableIntegrityHashes() specify the crossorigin attribute value (default: anonymous)
```
Lyrkan: Also, do we really want to only add it when the `integrity` attribute is also present? Couldn't it make sense to also need the `crossorigin` attribute for other reasons?

src/DependencyInjection/Configuration.php
PhilETaylor: Correct. It is possible to have the assets on the same domain name as your site, and therefore integrity can be set without a crossorigin attribute. I have now made changes to support this by allowing the default/false to not show the attribute.

src/Asset/TagRenderer.php
PhilETaylor: Yes. Done.

README.md
PhilETaylor: Correct. It is possible to have the assets on the same domain name as your site, and therefore integrity can be set without a crossorigin attribute. I have now made changes to support this by allowing the default/false to not show the attribute.

PhilETaylor:
@Lyrkan all your feedback has now been applied. Ready for testing/feedback/merging again now

PhilETaylor:
flex recipe also updated here symfony/recipes#562

src/Asset/TagRenderer.php
```diff
@@ -54,6 +58,10 @@ public function renderWebpackScriptTags(string $entryName, string $packageName =
                 $attributes['integrity'] = $integrityHashes[$filename];
             }
 
+            if (isset($this->defaultAttributes['crossorigin']) && false !== $this->defaultAttributes['crossorigin']) {
```
stof: this does not make sense. If you take an array `$defaultAttributes` in the renderer, you should merge them in attributes, not require implementing support for them one by one.

src/DependencyInjection/WebpackEncoreExtension.php
```diff
@@ -51,6 +51,11 @@ public function load(array $configs, ContainerBuilder $container)
         $container->getDefinition('webpack_encore.entrypoint_lookup_collection')
             ->replaceArgument(0, ServiceLocatorTagPass::register($container, $factories));
         $container->setAlias(EntrypointLookupInterface::class, new Alias($this->getEntrypointServiceId('_default')));
+
+        $container->getDefinition('webpack_encore.tag_renderer')
+            ->replaceArgument(2, [
+                'crossorigin' => $config['crossorigin'],
```
stof: I would rather filter out `false` value here at compile time rather than filtering them at runtime.

src/Asset/TagRenderer.php
PhilETaylor: ok thats probably beyond my symfony skill level - sorry. Happy for you to propose the code, show me an example, or take over this?

stof: Well, instead of `$attributes = [ 'src' => ... ]` at the beginning, make it like that:

```php
$attributes = $this->defaultAttributes;
$attributes['src'] = ...
```

there is nothing specific to Symfony here.

stof: note that this snippet assumes that `$this->defaultAttributes` contains only attributes which should be applied, not keys with `false` values when omitting them.

PhilETaylor: ah I thought you meant something with an options resolver :) 

ok I'll give it another go :)

PhilETaylor: ok committed a change to incorporate your feedback. ok?

src/DependencyInjection/WebpackEncoreExtension.php
PhilETaylor: I think I have implemented this like you recommend. ok?

src/Resources/config/services.xml
```diff
@@ -16,6 +16,9 @@
         <service id="webpack_encore.tag_renderer" class="Symfony\WebpackEncoreBundle\Asset\TagRenderer">
             <argument type="service" id="webpack_encore.entrypoint_lookup_collection" />
             <argument type="service" id="assets.packages" />
+            <argument type="collection">
+                <argument key="crossorigin" />
```
stof: this one should be removed. The default config should be an empty array (so `<argument type="collection" />`)

stof (changes requested):

src/Resources/config/services.xml
PhilETaylor: Done.

Lyrkan (commented):
Looks good to me :)

Maybe we're just missing some test cases for the default and explicit `false` values?

PhilETaylor:
In 7eeb941 I have reverted (and renamed) one of the tests so that it renders with no configured attributes, this is a quick win and covers the [TagRenderer](https://github.com/symfony/webpack-encore-bundle/blob/master/src/Asset/TagRenderer.php) class only, I guess what is really needed is some coverage of the [WebpackEncoreExtension](https://github.com/symfony/webpack-encore-bundle/blob/master/src/DependencyInjection/WebpackEncoreExtension.php) final class to check if the configuration attributes resolve correctly but there was no existing test and im not in the mood for that challenge today, sorry :) :)

I have also notices that LOADS of DocBlocks are missing on methods in this bundle - any ideas if that is on purpose or is that another quick win we could make using phpStorm to generate these missing DocBlocks?

Lyrkan:
>  I guess what is really needed is some coverage of the WebpackEncoreExtension final class to check if the configuration attributes resolve correctly but there was no existing test

The extension is tested by the [`IntegrationTest`](https://github.com/symfony/webpack-encore-bundle/blob/master/tests/IntegrationTest.php):

https://github.com/symfony/webpack-encore-bundle/blob/fc18456ed44737995f912c2084505fb67d040e11/tests/IntegrationTest.php#L147-L153

> I have also notices that LOADS of DocBlocks are missing on methods in this bundle - any ideas if that is on purpose or is that another quick win we could make using phpStorm to generate these missing DocBlocks?

Do you have an example of what you'd add? Most methods are already typed (params and return) so it wouldn't be really useful to add the same thing using PHPDoc.

Anyway, this should probably be in a separate PR to avoid mixing those changes with the ones here :)

PhilETaylor:
I'm just not used to seeing code without the additional PHPDoc for every method :-) I'll ignore that for now, phpStorm is great at reading the typehints so doesnt need the documentation to build its index anyway

I have added the default crossorigin = false in the IntegrationTest, the tests pass with it set to false, or not there at all. 

I guess this is ready for merging ?

Lyrkan (approved):

weaverryan (approved):

weaverryan:
Thanks for researching and fixing this Phil - excellent PR!

PhilETaylor:
🎉

PR #59, opened 2019-04-04
Only add alias to EntrypointLookupInterface if default build is enabled

emodric:
Caused by the combination of https://github.com/symfony/webpack-encore-bundle/commit/64f1e94140cbcc58beb4199ddc9633d8be15a4e2 and #37

The commit in question was added after the commits in my PR #37, so it slipped through the cracks.


weaverryan:
Thanks Edi!

PR #61, opened 2019-04-11
New preload option to automatically set preload Link header for rendered scripts/styles

### LINKED ISSUE #14: Investigate the way to render preload links easily
Currently, the bundle performs de-duplication of the returned files. This is useful to avoid rendering the same `<script>` tag multiple times. But it makes it harder to render `<link rel="preload">` in the `<head>` and then `<script>` at the end of the body (as here, we precisely need to get the full list twice).

The way to do it currently would be to call `encore_entry_js_files` and store the result in a variable used in both places. But this is not convenient at all. The variable must be defined in a scope higher than both places needing it, and it is very common to have a block wrapping the `<script>` tags to make it handy for child templates to add more, but blocks create a new scope and so mutating the array inside the block has no effect outside it. It is also quite fragile in case the variable name is used elsewhere in the template (as the head and the end of the body are at opposite sides of the rendering of the template).

weaverryan:
Fixes #14

This adds a new configuration option:

```yml
webpack_encore:
    # ...

    preload: true
```

When enabled, it automatically adds a `preload` entry to the Link header for
every JavaScript and CSS file that was rendered - for example:

Screen Shot 2019-04-11 at 11 51 36 AM

What I need is just some validation on this approach and that the Link header looks right ;).

Cheers!


composer.json
```diff
@@ -32,7 +32,8 @@
         "symfony/framework-bundle": "^3.4 || ^4.0",
         "symfony/phpunit-bridge": "^3.4 || ^4.1",
         "symfony/twig-bundle": "^3.4 || ^4.0",
-        "twig/twig": "^1.35 || ^2.0"
+        "twig/twig": "^1.35 || ^2.0",
+        "symfony/web-link": "^3.4 || ^4.0"
```
Lyrkan: Shouldn't `fig/link-util` also be required since `GenericLinkProvider` and `Link` are directly used? Or is a transitive dependency acceptable in this case?

PR #71, opened 2019-06-19
Autowire defaultBuild argument when it is not disabled

### LINKED ISSUE #70: Missing argument 2 (default build name) in webpack_encore.entrypoint_lookup_collection
For the `webpack_encore.entrypoint_lookup_collection` service, we're currently NEVER passing the 2nd argument: `$defaultBuildName`. In `WebpackEncoreExtension`, we should set this argument to `_default` inside this if statement: https://github.com/symfony/webpack-encore-bundle/blob/767a2159cce860da3278649b0ee5ce460758319c/src/DependencyInjection/WebpackEncoreExtension.php#L39-L42

MolloKhan:
Closes #70


weaverryan:
Thanks Diego!

PR #72, opened 2019-06-28
Bugfix: Properly add tags to services

Nyholm:
We need to make sure we can reset all resettable services.


weaverryan:
Thank you Tobias!

Nyholm:
Thank you for merging.

B-Galati:
Could be a good idea to update tests as well no?

Issue #73, opened 2019-07-09
Rendering template twice causes encore_entry_link_tags() to output nothing second time


andyexeter:
When calling `encore_entry_link_tags()` in a Twig template, the underlying PHP code has some safe-guarding in place to ensure the same link tag isn't outputted twice:

https://github.com/symfony/webpack-encore-bundle/blob/5e1cab3d223f65933d59a5a95ea01a6ed2833db4/src/Asset/EntrypointLookup.php#L84-L89

I'm not entirely sure of the rationale behind this, but it causes an issue when `encore_entry_link_tags()` is called twice in the same request.

This issue came to light when investigating why an order confirmation page of a website wasn't outputting a stylesheet's `link` tag. After some debugging I figured out it was because the controller for this page sends a confirmation email with an attached PDF which is rendered via HTML/Twig.

The Twig template for the PDF called `encore_entry_link_tags()`, so when the confirmation page was loaded and `encore_entry_link_tags()` was called a second time, nothing was returned.

I have worked around this by calling the `reset()` method of the `EntrypointLookup` object right after the confirmation email is sent but it feels like a bit of a hack.

Why does the "make sure to not return the same file multiple times" code need to be there? Can it be removed, or can the issue it solves be solved in a different way?

Relevant Slack discussion: https://symfony-devs.slack.com/archives/C5VHNHY11/p1562584991057800

Lyrkan:
Hey @andyexeter,

> Why does the "make sure to not return the same file multiple times" code need to be there?

That's because entrypoints can share some chunks.

For instance if you have `app.css`, `admin.css` and `app~admin.css` (common chunk) you may also want to call `{{ encore_entry_link_tags('app') }}` and `{{ encore_entry_link_tags('admin') }}` on the same page.
If that check didn't exist you'd then end-up with the link tag for `app~admin.css` being added multiple times to your page (the same thing applies to `encore_entry_script_tags`).

> Can it be removed, or can the issue it solves be solved in a different way?

It's a tricky topic... that behavior also causes issues in some cases such as error pages generation since that can happen *after* you already started processing your main template (see https://github.com/symfony/webpack-encore-bundle/pull/21)

weaverryan:
What we “kinda” want to do is making this “memory” this has almost tied to each Twig “rendering cycle”. Like, each time you render Twig, you get a fresh “memory/cache” that’s shared by all rendered sub-templates. But I’m not sure if knowing such a thing is even possible.

[...]

PR #74, opened 2019-08-06
Reset default EntrypointLookup on exception to fix #21 and #73

### LINKED ISSUE #910: Tricky error related to 500 error pages

tbmatuka:
Since @ckrack seems to be unavailable to continue working on #21, I figured this would be faster :)

I don't like having `_default` hardcoded in the listener, but I see no other options right now.
I thought about adding a `resetAll()` method to EntrypointLookupCollection, but there were a couple of issues with that:
1. It would either be a BC break if I added it to the interface, or an important method that isn't interfaced and I don't really like either of those.
2. I couldn't even implement it because the container that we get in the collection only has `has()` and `get()` methods, so I couldn't go through it. This would also have to be replaced (and break BC) to implement `resetAll()`.

Fixes https://github.com/symfony/demo/issues/910

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of the Symfony WebpackEncoreBundle package.
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\WebpackEncoreBundle\EventListener;
+
+use Symfony\WebpackEncoreBundle\Asset\EntrypointLookupCollection;
+
+class ExceptionListener
+{
+    private $entrypointLookupCollection;
+
+    private $buildNames;
+
+    public function __construct(EntrypointLookupCollection $entrypointLookupCollection, array $buildNames)
+    {
+        $this->entrypointLookupCollection = $entrypointLookupCollection;
+        $this->buildNames = $buildNames;
+    }
+
+    public function onKernelException()
+    {
+        foreach ($this->buildNames as $buildName) {
+            $this->entrypointLookupCollection->getEntrypointLookup($buildName)->reset();
+        }
+    }
+}
--- a/src/Resources/config/services.xml
+++ b/src/Resources/config/services.xml
@@ -50,6 +50,12 @@
             <tag name="cache.pool" />
         </service>
 
+        <service id="webpack_encore.exception_listener" class="Symfony\WebpackEncoreBundle\EventListener\ExceptionListener">
+            <tag name="kernel.event_listener" event="kernel.exception" />
+            <argument type="service" id="webpack_encore.entrypoint_lookup_collection" />
+            <argument /> <!-- build list of build names -->
+        </service>
+
         <service id="webpack_encore.preload_assets_event_listener" class="Symfony\WebpackEncoreBundle\EventListener\PreLoadAssetsEventListener">
             <tag name="kernel.event_subscriber" />
             <argument type="service" id="webpack_encore.tag_renderer" />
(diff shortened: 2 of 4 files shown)


Lyrkan:
> I don't like having _default hardcoded in the listener, but I see no other options right now.
> I thought about adding a resetAll() method to EntrypointLookupCollection, but there were a couple of issues with that

Wouldn't it be possible to handle that using the `WebpackEncoreExtension` (even though I agree that having a `all()` and/or a `resetAll()` in the `EntrypointLookupCollection` would have been a better solution)?

1) add a `$builds` argument to the constructor of `webpack_encore.exception_listener`
2) use `$container->getDefinition('webpack_encore.exception_listener')->replaceArgument(...)` to inject the list of available builds in it

tbmatuka:
I made it work like you suggested. I still don't love it, but I like it a lot better than before.

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,26 @@
...
+namespace Symfony\WebpackEncoreBundle\EventListener;
+
+use Symfony\WebpackEncoreBundle\Asset\EntrypointLookup;
+use Symfony\WebpackEncoreBundle\Asset\EntrypointLookupCollection;
+
+class ExceptionListener
+{
+    private $entrypointLookupCollection;
+
+    private $entrypointNames;
```
weaverryan: I believe these are `buildNames` - we should change `entrypoint` -> `build` everywhere in this class

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Symfony\WebpackEncoreBundle\EventListener;
```
weaverryan, line 10: can you add LICENSE header at the top?

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,26 @@
...
+        $this->entrypointNames = $entrypointNames;
+    }
+
+    public function onKernelException()
+    {
+        foreach ($this->entrypointNames as $entrypointName) {
+            $this->entrypointLookupCollection->getEntrypointLookup($entrypointName)->reset();
+        }
+    }
+}
```
weaverryan, line 32: Could we add a very basic unit test for this class? There isn't much to test, but having at least ONE basic test will protect us from making any super silly typos now or in the future.

src/Resources/config/services.xml
```diff
@@ -50,6 +50,12 @@
             <tag name="cache.pool" />
         </service>
 
+        <service id="webpack_encore.exception_listener" class="Symfony\WebpackEncoreBundle\EventListener\ExceptionListener">
+            <tag name="kernel.event_listener" event="kernel.exception" />
+            <argument type="service" id="webpack_encore.entrypoint_lookup_collection" />
+            <argument /> <!-- build list of entrypoint paths -->
```
weaverryan:
```suggestion
            <argument /> <!--collection of build names -->
```

weaverryan (changes requested):
Minor comments - thanks for moving this forward!

tbmatuka:
@weaverryan should be all done, can you take a look?

tbmatuka:
@weaverryan can you please take a look?

tbmatuka:
@weaverryan it's been a month now, can you please take a quick look so that we can move this along?

tbmatuka:
@weaverryan @Lyrkan this has been waiting for review/merge almost three months now, can we please move it along?

weaverryan:
Sorry for the long delay @tbmatuka - but thank you SO much for finishing this :). Cheers!

PR #78, opened 2019-08-27
Fix Travis SYMFONY_REQUIRE config

wouterj:
The `composer install` script was running in `before_install`, while the `SYMFONY_REQUIRE` + symfony/flex was configured in `install`. Swapping these two from position fixes the Travis setup. (flex has to be configured before `composer install`).

I also removed `if ! [ -v "$DEPENDENCIES" ]; then composer require --no-update ${DEPENDENCIES}; fi;`, it seems to be a left over from previously used configs as the `DEPENDENCIES` is not defined anywhere.


nicolas-grekas (approved):

maxhelias (approved):

yceruto (approved):

nicolas-grekas:
Thank you @wouterj.

PR #80, opened 2019-10-22
Allow Symfony 5

nicolas-grekas:


weaverryan:
Thank you Nicolas!

composer.json
```diff
@@ -19,21 +19,20 @@
             "Symfony\\WebpackEncoreBundle\\Tests\\": "tests/"
         }
     },
+    "minimum-stability": "dev",
```
B-Galati, line 22: Did this stay on purpose ?

nicolas-grekas: Yes, it's a best practice to do this for libs.

B-Galati: Thanks! I guess it's because it allows to install unstable deps easily when working on the bundle?

PR #91, opened 2020-04-24
Add new functions to disable "file tracking" & ease inline_css usage

### LINKED ISSUE #90: Ability to "reset" cache of rendered files automatically in Twig
Related #87 

When you `encore_entry_link_tags()` or `encore_entry_script_tags()` tags, the bundle "remembers" which CSS and JS files have been rendered and avoids rendering them again. This is by design: it allows you to render multiple "entries" on the same page and avoid link/script tags for the same CSS & JS files (due to code splitting).

But, this has some nasty side effects. One practical one is sending emails and using these functions to import your CSS. If you have Messenger installed, each Email is technically rendered twice - and the second (real) email will contain *no* CSS link tags. The problem is probably worse when sending emails through the queue - as I think once a worker has sent one email, the rest won't contain CSS link tags.

The best solution would be to reset "automatically" the cache after a "main" Twig rendering (e.g. a call to`$twig->render()`) but no such concept currently exists in Twig.

A more immediate solution may be to allow users to "opt out" of the cache via some argument to `encore_entry_link_tags()`.

weaverryan:
Hi!

Fixes #90 

This contains several new features to help integration primarily with Mailer:

## encore_disable_file_tracking()
```twig
{% do encore_disable_file_tracking() %}
    {{ encore_entry_link_tags('entry1') }}
    {{ encore_entry_script_tags('entry1') }}
{% do encore_enable_file_tracking() %}
```

This is for usage in Email or PDF templates to tell encore to "not" use its "file tracking" system temporarily. This avoids issues where, for example, if you rendered 2 emails at once, the 2nd would have no link tags because Encore things they are already rendered.

I also considered an alternative syntax - just having `{% encore_encore_file_tracking %}` one time at the top of the template. If people really like that, we can try that.

## 2) encore_entry_css_source()

This file has "tracking" automatically disabled.

```twig
{% apply inline_css(encore_entry_css_source('my_entry')) %}
    <div>
        Hi! The CSS from my_entry will be converted into
        inline styles on any HTML elements inside.
    </div>
{% endapply %}
```

I'd appreciate any feedback! Thanks!

Note to self: will simplify https://symfonycasts.com/screencast/mailer/pdf-styles and https://symfonycasts.com/screencast/mailer/encore-inline_css

### DIFF
--- a/tests/Asset/BuildFileLocatorTest.php
+++ b/tests/Asset/BuildFileLocatorTest.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of the Symfony WebpackEncoreBundle package.
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\WebpackEncoreBundle\Tests\Asset;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\WebpackEncoreBundle\Asset\BuildFileLocator;
+
+class BuildFileLocatorTest extends TestCase
+{
+    /**
+     * @dataProvider getFindFileTests
+     */
+    public function testFindFile(string $buildPath, string $assetPath, string $expectedPath)
+    {
+        $fileLocator = new BuildFileLocator(['_default' => $buildPath]);
+        $fileLocator->disableFileExistsCheck();
+
+        $this->assertSame($expectedPath, $fileLocator->findFile($assetPath));
+    }
+
+    public function getFindFileTests()
+    {
+        yield 'no_overlap' => [
+            '/app/public',
+            'foo.js',
+            '/app/public/foo.js'
+        ];
+
+        yield 'simple_overlap' => [
+            '/app/public/build',
+            'build/foo.js',
+            '/app/public/build/foo.js'
+        ];
+
+        yield 'simple_overlap_with_slash' => [
+            '/app/public/build',
+            '/build/foo.js',
+            '/app/public/build/foo.js'
+        ];
+
+        yield 'simple_overlap_with_build_path_slash' => [
+            '/app/public/build/',
+            'build/foo.js',
+            '/app/public/build/foo.js'
+        ];
+
+        yield 'partial_overlap' => [
+            '/app/public/build',
+            'build/subdirectory/foo.js',
+            '/app/public/build/subdirectory/foo.js'
+        ];
+
+        yield 'overlap_in_wrong_spot' => [
+            '/app/public/build',
+            'subdirectory/build/foo.js',
+            '/app/public/build/subdirectory/build/foo.js'
+        ];
+
+        yield 'windows_paths' => [
+            'C:\\\\app\\public\\build',
+            'build/foo.js',
+            'C:\\\\app\\public\\build/build/foo.js'
+        ];
+    }
+
+    // test only CSS & JS allowed
+}
--- a/README.md
+++ b/README.md
@@ -105,21 +105,67 @@ If you want more control, you can use the `encore_entry_js_files()` and
 `encore_entry_css_files()` methods to get the list of files needed, then
 loop and create the `script` and `link` tags manually.
 
-## Rendering Multiple Times in a Request (e.g. to Generate a PDF)
+## Rendering Multiple Templates (e.g. Emails or PDFs)
 
 When you render your script or link tags, the bundle is smart enough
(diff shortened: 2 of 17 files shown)


javiereguiluz:
Thanks for this contribution!

I'm a bit worried because of the added complexity and because the new feature is not obvious to discover.

I don't know if there are better names ... but "disable file tracking" it's a bit confusing because it's the indirect solution to the problem users are facing. If you have this issue, you'll Google for "encore assets disappear", "encore assets not available in the template", etc.  The "file tracking" is the indirect cause of the problem ... that's why it's hard to associate both.

The general problem is that for "setter operations", this is very easy:

* `set()` = set and replace existing
* `add()` = set but not replace existing

For "getters operations":

* `get()` = get and keep it (<-- this is the proposed `encore_disable_file_tracking()`)
* `.....()` = get and consume it (<-- this is the current behavior of Encore)

I don't know if there's a standard name in the industry for "get and consume". I asked in Symfony Slack and they proposed `eat()`, `consume()`, `pop()`, etc.

tests/Asset/BuildFileLocatorTest.php
```diff
@@ -0,0 +1,74 @@
...
+            '/app/public/build',
+            'subdirectory/build/foo.js',
+            '/app/public/build/subdirectory/build/foo.js'
+        ];
+
+        yield 'windows_paths' => [
+            'C:\\\\app\\public\\build',
+            'build/foo.js',
+            'C:\\\\app\\public\\build/build/foo.js'
+        ];
```
Kocal, line 70: Are files from dev-server supported?

Because that's an issue my team and me faced when working with Encore dev-server and assets in Twig for emails (with `inky_to_html` and `inline_css`)

We were able to make them work with a custom extension:

```php
class EncoreExtension extends AbstractExtension
{
    private $publicDir;

    public function __construct(string $publicDir)
    {
        $this->publicDir = $publicDir;
    }

    public function getFunctions(): array
    {
        return [
            new TwigFunction('encore_source', [$this, 'source']),
        ];
    }

    public function source(string $path): string
    {
        // don't do anything if $path is http://my-app.domain:8008/...
        if (0 === strpos($path, '/')) {
            $path = $this->publicDir.$path;
        }

        // fetch content from filesystem or from http
        if (false !== $content = file_get_contents($path)) {
            return $content;
        }

        throw new \RuntimeException("Unable to read file \"$path\".");
    }
}
```

```twig
{% set css_to_inline = encore_entry_css_files('email') | map(file => encore_source(file)) | join('') %}
{% set css_to_ship = encore_entry_css_files('email-responsive') | map(file => encore_source(file)) | join('') %}

{% apply inky_to_html|inline_css(css_to_inline) %}
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width">

    <style>
    {{ css_to_ship }}
    </style>

    {# ... #}
```

But that would be really nice if it can be supported _natively_ here :D

weaverryan: That's a really good point - I hadn't forgotten about the dev server. Assuming it *did* support the dev server (it actually does not currently), would the new `encore_entry_css_source()` Twig function work well for you?

Kocal: I will give it a try!

Kocal: I've updated my code to this:
```diff
-{% set css_to_inline = encore_entry_css_files('email') | map(file => encore_source(file)) | join('') %}
-{% set css_to_ship = encore_entry_css_files('email-responsive') | map(file => encore_source(file)) | join('') %}
- 
-{% apply inky_to_html|inline_css(css_to_inline) %}
+{% apply inky_to_html|inline_css(encore_entry_css_source('email')) %}
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width">

    <style>
-    {{ css_to_ship }}
+    {{ encore_entry_css_source('email-responsive') }}
    </style>

    <style>
```

It works when running `encore dev`, but not when running the dev-server:

Capture d’écran de 2020-05-09 00-42-17

stof:
Rather than a function to call in a `do` tag before and after (for which it is easy to forget undoing it after), I think this should be a tag instead (which will reset the state when reaching the end tag)

Warxcell:
Instead of manually handling this - I think it should be automatically done. 

https://github.com/symfony/webpack-encore-bundle/pull/115

javiereguiluz:
I think we should close this PR as "won't fix". We're focusing exclusively on AssetMapper, so we should no longer add new features to Webpack Encore. Thanks.

PR #97, opened 2020-10-08
Allow entrypoints.json to be hosted remotely

### LINKED ISSUE #76: Allow output_path to be remote url
We have a case where our css and js files are on a cdn and we have encore build all the files remotely on the cdn after we do a push. the entrypoints.json is therefore also remotely on the cdn server.

I can add this in the config:

webpack_encore:
    output_path: "https://cdn.blabla.com/experimental"

But then the file_exists check in src/Asset/EntryPointLookup.php  on line 120 halts the process.
Whereas the file_get_contents on line 127 would work perfectly well with the remote url.

So could there be a way to bypass the file_exists check if the output_path is a url? Or add a config option "remote" or something like that?

rlvdx:
Fixes #76 

- When the assets Webpack Encore builds are not hosted on a Docker container's filesystem but published remotely (on a Google Cloud Storage bucket for example), `file_exists` on `entrypoints.json` fails.
- When `strict_mode` is `true`, an exception is raised and it's imposible to load the assets.
- Just after this check and condition, a `file_get_contents` is used to retrieve the `entrypoints.json` contents.

What I'm proposing is to use `file_get_contents` earlier to allow fetching a remote `entrypoints.json` to solve the CDN use case. Does it seem OK?

On a side note, I don't see how to run the tests locally, so the feedback loop with TRavis makes it difficult to fix them. How should I do? Thanks :)

### DIFF
--- a/src/Asset/EntrypointLookup.php
+++ b/src/Asset/EntrypointLookup.php
@@ -117,14 +117,15 @@ class EntrypointLookup implements EntrypointLookupInterface, IntegrityDataProvid
             }
         }
 
-        if (!file_exists($this->entrypointJsonPath)) {
+        $entrypointJsonContents = file_get_contents($this->entrypointJsonPath);
+        if ($entrypointJsonContents === false) {
             if (!$this->strictMode) {
                 return [];
             }
-            throw new \InvalidArgumentException(sprintf('Could not find the entrypoints file from Webpack: the file "%s" does not exist.', $this->entrypointJsonPath));
+            throw new \InvalidArgumentException(sprintf('Could not find the entrypoints file from Webpack: the file "%s" does not exist or it is not readable.', $this->entrypointJsonPath));
         }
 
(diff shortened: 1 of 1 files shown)


piotrantosik:
hi, any news? it's any plan to implement this feature?

rlvdx:
> hi, any news? it's any plan to implement this feature?

Doesn't look like it, and I don't know how to make things go forward :)

roublarstar:
Hello, I am in the same situation as you ... your solution seems viable to me too!

Ahummeling (approved):
Seems exactly like what's needed.
I'm not a maintainer of this library, but it would be nice if we could have this.

Kocal:
I'm on it :)

Kocal:
Sorry but I wasn't able to push on your PR, instead I've opened a new PR #236

Kocal:
Released in https://github.com/symfony/webpack-encore-bundle/releases/tag/v2.2.0

PR #98, opened 2020-10-20
PHP 8.0 compatibility

jmsche:
Tests are also run against PHP7.4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,14 @@ matrix:
         - php: 7.2
         - php: 7.3
           env: deps=low
+        - php: 7.4
+          env: SYMFONY_PHPUNIT_VERSION=9.4
+        - php: nightly
+          env: SYMFONY_PHPUNIT_VERSION=9.4
 
 before_install:
     - phpenv config-rm xdebug.ini || true
-    - composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-master
+    - composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-main
 
 install:
     - |
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     },
     "minimum-stability": "dev",
     "require": {
-        "php": "^7.1.3",
+        "php": ">=7.1.3",
         "symfony/asset": "^3.4 || ^4.0 || ^5.0",
         "symfony/config": "^3.4 || ^4.0 || ^5.0",
         "symfony/dependency-injection": "^3.4 || ^4.0 || ^5.0",
--- a/tests/Asset/TagRendererTest.php
+++ b/tests/Asset/TagRendererTest.php
@@ -13,8 +13,8 @@ use PHPUnit\Framework\TestCase;
 use Symfony\Component\Asset\Packages;
 use Symfony\WebpackEncoreBundle\Asset\EntrypointLookupCollection;
 use Symfony\WebpackEncoreBundle\Asset\EntrypointLookupInterface;
-use Symfony\WebpackEncoreBundle\Asset\IntegrityDataProviderInterface;
 use Symfony\WebpackEncoreBundle\Asset\TagRenderer;
+use Symfony\WebpackEncoreBundle\Tests\TestEntrypointLookupIntegrityDataProviderInterface;
 
 class TagRendererTest extends TestCase
 {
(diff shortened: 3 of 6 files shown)


composer.json
```diff
@@ -21,7 +21,7 @@
     },
     "minimum-stability": "dev",
     "require": {
-        "php": "^7.1.3",
+        "php": "^7.1.3 || ^8",
```
maxhelias: \>=7.1.3 ? Like the other dependencies of Symfony?

jmsche: Fixed, thank you :)

jmsche:
So, the tests pass, but Travis fails because of this deprecation:

> Since symfony/twig-bundle 5.2: Accessing the "twig" service directly from the container is deprecated, use dependency injection instead.

However I can't find a way to get Twig Environment properly from the container, any idea about this?

Kocal:
You can inject `\Twig\Environment` in your constructor.

maxhelias:
@Kocal Not so easy for testing

@jmsche I'm not sure, but take a look at the class `Symfony\Bundle\FrameworkBundle\Test\KernelTestCase` and the service `test.service_container`. It might take a little work but I think it's a good lead.

Kocal:
@maxhelias yes that's not really easy in fact, I'm already helping @jmsche on Symfony Slack and I think we will use the `test.service_container`.

jmsche:
Finally Travis is green :)

Many thanks to @Kocal who helped me a lot :)

maxhelias (approved):
Look good 👍

tests/EntrypointLookupIntegrityDataProviderInterface.php
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+namespace Symfony\WebpackEncoreBundle\Tests;
+
+use Symfony\WebpackEncoreBundle\Asset\EntrypointLookupInterface;
+use Symfony\WebpackEncoreBundle\Asset\IntegrityDataProviderInterface;
+
+interface EntrypointLookupIntegrityDataProviderInterface extends EntrypointLookupInterface, IntegrityDataProviderInterface
+{
+}
```
weaverryan: Can you explain why this new interface is needed? Just for convenience? To avoid a problem with using `$this->createMock()` and passing an array? Something else?

weaverryan (commented):
Just one small comment - I think I might be missing a detail. Thanks for this - it looks great!

tests/EntrypointLookupIntegrityDataProviderInterface.php
jmsche: This new interface exists because createMock() only accepts a string as parameter since phpunit 9.0: https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestCase.php#L1698

.travis.yml
```diff
@@ -18,6 +18,10 @@ matrix:
         - php: 7.2
         - php: 7.3
           env: deps=low
+        - php: 7.4
+          env: SYMFONY_PHPUNIT_VERSION=9.4
+        - php: nightly
+          env: SYMFONY_PHPUNIT_VERSION=9.4
```
jrushlow, line 24: Why PHPUnit 9.4 specifically?

jmsche: I chose 9.4 as it's the most recent right now, so the most susceptible one to have bug fixes.

Requires at least 9.3 as support for PHP 8 was added at that time.

tests/EntrypointLookupIntegrityDataProviderInterface.php
jrushlow: Because this is only used in the `TagRendererTest::class` - I would remove the interface here and just add it at the end of the `TagRendererTest.php` file. Using the signature `TestEntrypointLookupIntegrityDataProviderInterface::class` for the interface would also help mitigate someone from accidentally using this interface outside of the test suite.

jmsche: From my POV the tests/ directory (and other files) should be excluded from the release package using .gitattributes.  
Should be done in an other PR though.

About moving the interface directly in the class file, I'm generally against having multiple class/interface/etc declarations in the same file, but I can do it if you insist :)

weaverryan: Let's rename it to `TestEntrypointLookupIntegrityDataProviderInterface`. We can ignore the `tests/` directory in another PR, but I'd still prefer that classes for tests *look* like test classes.

Thanks!

jmsche: Done! :)

jmsche:
Updated PR by renaming test interface & fixing symfony/flex branch name in Travis config.

Travis is green again! :)

jrushlow (approved):

weaverryan:
Thank you so much for this detailed work @jmsche!

PR #102, opened 2021-01-15
Adding support for custom attributes on rendered script and link tags

### LINKED ISSUE #10: Allow custom script/link tag attributes to be passed

weaverryan:
Fixes #10 - thanks to @dsech for the idea of relying on named Twig arguments (otherwise the new Twig attributes argument is the 4th argument... and rather inconvenient).

This can be configured in *3* different ways:

## 1) Globally

```
webpack_encore:
    # ...
    script_attributes:
        defer: true
```

2) when when using the Twig functions:

```twig
{{ encore_entry_script_tags('my_entry', attributes={
    defer: true
}) }}
```

3) Via a new `RenderAssetTagEvent`

As stated at the bottom of #10 - https://github.com/symfony/webpack-encore-bundle/issues/10#issuecomment-756835238 - there may be some additional options that would also be nice, but this probably covers the biggest use-case.

NOTE: This also drops support for EOL Symfony versions.

Cheers!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,17 +22,17 @@
     "minimum-stability": "dev",
     "require": {
         "php": ">=7.1.3",
-        "symfony/asset": "^3.4 || ^4.0 || ^5.0",
-        "symfony/config": "^3.4 || ^4.0 || ^5.0",
-        "symfony/dependency-injection": "^3.4 || ^4.0 || ^5.0",
-        "symfony/http-kernel": "^3.4 || ^4.0 || ^5.0",
+        "symfony/asset": "^4.4 || ^5.0",
+        "symfony/config": "^4.4 || ^5.0",
+        "symfony/dependency-injection": "^4.4 || ^5.0",
+        "symfony/http-kernel": "^4.4 || ^5.0",
         "symfony/service-contracts": "^1.0 || ^2.0"
     },
     "require-dev": {
-        "symfony/framework-bundle": "^3.4 || ^4.0 || ^5.0",
-        "symfony/phpunit-bridge": "^4.3.5 || ^5.0",
-        "symfony/twig-bundle": "^3.4 || ^4.0 || ^5.0",
-        "symfony/web-link": "^3.4 || ^4.0 || ^5.0"
+        "symfony/framework-bundle": "^4.4 || ^5.0",
+        "symfony/phpunit-bridge": "^4.4 || ^5.0",
+        "symfony/twig-bundle": "^4.4 || ^5.0",
+        "symfony/web-link": "^4.4 || ^5.0"
     },
     "extra": {
         "thanks": {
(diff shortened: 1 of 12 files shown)


src/Asset/TagRenderer.php
```diff
@@ -47,24 +55,43 @@ public function __construct(
...
         $this->packages = $packages;
         $this->defaultAttributes = $defaultAttributes;
+        $this->defaultScriptAttributes = $defaultScriptAttributes;
+        $this->defaultLinkAttributes = $defaultLinkAttributes;
 
         $this->reset();
     }
 
-    public function renderWebpackScriptTags(string $entryName, string $packageName = null, string $entrypointName = '_default'): string
+    public function setEventDispatcher(EventDispatcherInterface $eventDispatcher)
```
dunglas: Why using setter injection instead of an optional constructor argument?

src/Event/RenderAssetTagEvent.php
```diff
@@ -0,0 +1,64 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\WebpackEncoreBundle\Event;
+
+/**
+ * Dispatched each time a script or link tag is rendered.
+ */
+class RenderAssetTagEvent
```
dunglas: This class could maybe be marked `final`.

dunglas (approved):

composer.json
```diff
@@ -22,17 +22,17 @@
     "minimum-stability": "dev",
     "require": {
         "php": ">=7.1.3",
```
tgalopin, line 24: You can probably bump to 7.2 as Symfony 4.4 requires it. This would allow you to remove a CI job too :)

src/Asset/TagRenderer.php
weaverryan: it was a subjective call... it felt odd to put a service argument after 3 optional array args.

composer.json
weaverryan: I would love to - but I think 4.4 only requires 7.1 - https://github.com/symfony/skeleton/blob/1d596be50b63df6e8fdf94f89f7f38ae6d17e024/composer.json#L7

tgalopin: Ah indeed!

nicodemuz:
@weaverryan is it possible to add attributes before the `src` attribute?

This is a requirement to disable Cloudflare Rocket Loader for a particular script, see:
https://support.cloudflare.com/hc/en-us/articles/200169436-How-can-I-have-Rocket-Loader-ignore-specific-JavaScripts-

The above link mentions `The 'data-cfasync' attribute must be added before the 'src' attribute.`:
```
<script data-cfasync="false" src="/javascript.js"></script>
```

PR #104, opened 2021-01-25
Allow custom EntrypointLookupCollection when instantiating the TagRenderer

richardhj:
Hi 👋

There is an interface "EntrypointLookupCollectionInterface" but I wonder why the TagRender is not making use of it?

I want to override the TagRenderer to work with another EntrypointLookupCollection, so I defined my own service:

```yaml
    app.tag_renderer:
        class: Symfony\WebpackEncoreBundle\Asset\TagRenderer
        arguments:
            - '@app.entrypoint_lookup_collection'
            - '@assets.packages'
            - []
            - []
            - []
            - '@event_dispatcher'
        tags:
            - { name: 'kernel.reset', method: 'reset'}
```

### DIFF
--- a/src/Asset/TagRenderer.php
+++ b/src/Asset/TagRenderer.php
@@ -45,10 +45,10 @@ class TagRenderer implements ResetInterface
                     return $entrypointLookupCollection;
                 }])
             );
-        } elseif ($entrypointLookupCollection instanceof EntrypointLookupCollection) {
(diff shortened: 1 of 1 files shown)


richardhj:
Related: #2, #6, https://github.com/symfony/webpack-encore-bundle/pull/17/files#r234230374.

weaverryan:
Looks like this was a detail we missed - thanks Richard!

PR #109, opened 2021-02-09
Introduce stimulus_controller to ease Stimulus Values API usage

tgalopin:
This PR introduces a `stimulus_controller` helper to ease the usage of [Stimulus Values API](https://stimulus.hotwire.dev/reference/values).

The Values API is very useful as it allows developers to store state on the DOM node and thus persist it, share it across controllers and build reusable controllers. Using the Values API as a way to configure a controller using data attributes is a very common pattern that we should encourage ("build primitive reusable and composable controllers instead of building a controller per page").

However it has a few drawbacks:

1/ It introduces a duplication of the scoped attributes: 

```html
<div data-controller="symfony--ux-chartjs-chart"
    data-symfony--ux-chartjs-chart-data-value="..."
    data-symfony--ux-chartjs-chart-options-value="..."
....>
</div>
```

This can quickly become cumbersome and difficult to track.

2/ It creates its own convention on naming, if you declare a value using camel case, it will need to be passed using kebab case in the HTML:

```javascript
export default class extends Controller {
    static values = {
        myMessage: String,
    }

    connect() {
        this.element.textContent = this.myMessageValue;
    }
}
```

```html
<div data-controller="hello" data-hello-my-message-value="Hello"></div>
```

3/ It requires to manually encode nested structure in JSON in order to pass them to the controller

This PR improves upon these drawbacks: it introduces a helper you can use in your DOM node to render the controller to use, with optional values for this controller. These values will be encoded properly to be read by Stimulus natively (https://stimulus.hotwire.dev/reference/values#properties-and-attributes).

With a single controller:

```twig
<div {{ stimulus_controller({ 'chart': { 'data': [1, 2, 3, 4] } }) }}>
    Hello
</div>
```

With multiple controllers:

```twig
<div {{ stimulus_controller({
    'chart': {
        'data': [1, 2, 3, 4],
        'labels': ['January', 'February', 'March', 'April'],
    },
    'another-controller': {}
}) }}>
    Hello
</div>
```

It introduces a "standard" way to transfer data to Stimulus controllers that map well with the Values API and thus work out of the box.

### DIFF
--- a/src/Twig/StimulusTwigExtension.php
+++ b/src/Twig/StimulusTwigExtension.php
@@ -1,0 +1,77 @@
+<?php
+
+/*
+ * This file is part of the Symfony WebpackEncoreBundle package.
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\WebpackEncoreBundle\Twig;
+
+use Twig\Environment;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+final class StimulusTwigExtension extends AbstractExtension
+{
+    public function getFunctions()
+    {
+        return [
+            new TwigFunction('stimulus_controller', [$this, 'renderStimulusController'], ['needs_environment' => true, 'is_safe' => ['all']]),
+        ];
+    }
+
+    public function renderStimulusController(Environment $env, array $data): string
+    {
+        if (!$data) {
+            return '';
+        }
+
+        $controllers = [];
+        $values = [];
+
+        foreach ($data as $controllerName => $controllerValues) {
+            $controllerName = twig_escape_filter($env, $this->normalizeControllerName($controllerName), 'html_attr');
+            $controllers[] = $controllerName;
+
+            foreach ($controllerValues as $key => $value) {
+                if (!is_scalar($value)) {
+                    $value = json_encode($value);
+                }
+
+                $key = twig_escape_filter($env, $this->normalizeKeyName($key), 'html_attr');
+                $value = twig_escape_filter($env, $value, 'html_attr');
+
+                $values[] = 'data-'.$controllerName.'-'.$key.'-value="'.$value.'"';
+            }
+        }
+
+        return rtrim('data-controller="'.implode(' ', $controllers).'" '.implode(' ', $values));
+    }
+
+    /**
+     * Normalize a Stimulus controller name into its HTML equivalent (no special character and / becomes --).
+     *
+     * @see https://stimulus.hotwire.dev/reference/controllers
+     */
+    private function normalizeControllerName(string $str): string
+    {
+        return preg_replace('/^@/', '', str_replace('_', '-', str_replace('/', '--', $str)));
+    }
+
+    /**
+     * Normalize a Stimulus Value API key into its HTML equivalent ("kebab case").
+     * Backport features from symfony/string.
+     *
+     * @see https://stimulus.hotwire.dev/reference/values
+     */
+    private function normalizeKeyName(string $str): string
+    {
+        // Adapted from ByteString::camel
+        $str = ucfirst(str_replace(' ', '', ucwords(preg_replace('/[^a-zA-Z0-9\x7f-\xff]++/', ' ', $str))));
+
+        // Adapted from ByteString::snake
+        return strtolower(preg_replace(['/([A-Z]+)([A-Z][a-z])/', '/([a-z\d])([A-Z])/'], '\1-\2', $str));
+    }
+}
--- a/src/Resources/config/services.xml
+++ b/src/Resources/config/services.xml
@@ -36,6 +36,10 @@
             </argument>
         </service>
 
+        <service id="webpack_encore.twig_stimulus_extension" class="Symfony\WebpackEncoreBundle\Twig\StimulusTwigExtension">
+            <tag name="twig.extension" />
+        </service>
+
         <service id="webpack_encore.entrypoint_lookup.cache_warmer" class="Symfony\WebpackEncoreBundle\CacheWarmer\EntrypointCacheWarmer">
             <tag name="kernel.cache_warmer" />
             <argument /> <!-- build list of entrypoint paths -->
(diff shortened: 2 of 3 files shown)


src/Twig/StimulusTwigExtension.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        if (!$data) {
+            return '';
+        }
+
+        $controllers = [];
+        $values = [];
+
+        foreach ($data as $controllerName => $controllerValues) {
+            $controllers[] = $controllerName;
```
nicolas-grekas, line 36: `$controllerName` should be HTML escaped.

src/Twig/StimulusTwigExtension.php
```diff
@@ -0,0 +1,72 @@
...
+
+            foreach ($controllerValues as $key => $value) {
+                if (!is_scalar($value)) {
```
nicolas-grekas, line 39: why not scalars? shouldn't we always pass a json?

tgalopin: Not according to Stimulus: https://stimulus.hotwire.dev/reference/values#types

But perhaps json_encode always does the right thing?

tgalopin: Actually json_encode doesn't: true needs to be encoded in 1, false in 0.

nicolas-grekas: OK, the controller decides how it is going to parse the data!
Thanks for the link.

src/Twig/StimulusTwigExtension.php
```diff
@@ -0,0 +1,74 @@
...
+                }
+
+                $controllerName = twig_escape_filter($env, $controllerName, 'html_attr');
+                $value = twig_escape_filter($env, $value, 'html_attr');
+
+                $values[] = 'data-'.$controllerName.'-'.$this->convertKeyCase($key).'-value="'.$value.'"';
+            }
+        }
+
+        return rtrim('data-controller="'.implode(' ', $controllers).'" '.implode(' ', $values));
```
nicolas-grekas, line 50: `$controllers` contains unescaped names.

src/Twig/StimulusTwigExtension.php
tgalopin: Yes and that's to usefulness of the Values API: the parsing is done by Stimulus :)

src/Twig/StimulusTwigExtension.php
tgalopin: fixed

src/Twig/StimulusTwigExtension.php
```diff
@@ -0,0 +1,75 @@
...
+        foreach ($data as $controllerName => $controllerValues) {
+            $controllerName = twig_escape_filter($env, $controllerName, 'html_attr');
+            $controllers[] = $controllerName;
+
+            foreach ($controllerValues as $key => $value) {
+                if (!is_scalar($value)) {
+                    $value = json_encode($value);
+                }
+
+                $key = $this->convertKeyCase(twig_escape_filter($env, $key, 'html_attr'));
```
nicolas-grekas: escaping should happen after convertKeyCase, isn't it?

src/Twig/StimulusTwigExtension.php
```diff
@@ -0,0 +1,75 @@
...
+
+        return rtrim('data-controller="'.implode(' ', $controllers).'" '.implode(' ', $values));
+    }
+
+    /**
+     * Convert a Value key into the HTML equivalent for Stimulus ("kebab case").
+     * Backport features from symfony/string.
+     *
+     * @param string $string
+     * @return string
```
nicolas-grekas: can be removed

src/Twig/StimulusTwigExtension.php
tgalopin: I'm not sure :) . In practice this probably will never happen (you need to know the key, it can't really be user data), but just in case I thought if would be better to always have a formatted value for Stimulus not to break.

nicolas-grekas (approved):

src/Exception/MissingPackageException.php
```diff
@@ -0,0 +1,14 @@
...
+/*
+ * This file is part of the Symfony WebpackEncoreBundle package.
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\WebpackEncoreBundle\Exception;
+
+class MissingPackageException extends \LogicException
```
weaverryan: I think this is not used?

src/Twig/StimulusTwigExtension.php
```diff
@@ -0,0 +1,78 @@
...
+
+        return rtrim('data-controller="'.implode(' ', $controllers).'" '.implode(' ', $values));
+    }
+
+    /**
+     * Normalize a Stimulus controller name into its HTML equivalent (no special character and / becomes --).
+     *
+     * @see https://stimulus.hotwire.dev/reference/controllers
+     */
+    private function normalizeControllerName(string $str): string
```
weaverryan, line 58: I'm not sure this belongs here. It seems to me that the UX packages should be responsible for calling this function with their already-normalized name. I'd almost rather throw an exception if something invalid is passed (but probably, just do nothing).

weaverryan (commented):
I love this! Just minor comments/questions

src/Exception/MissingPackageException.php
tgalopin: Indeed, it was but isn't anymore!

src/Twig/StimulusTwigExtension.php
tgalopin: The idea was to allow users to pass the same value they have in their controller.json file: `@symfony/ux-dropzone/dropzone`, and this would be transformed to the proper HTML equivalent. But perhaps should we have an alternate way of doing this?

weaverryan: Hmm, that's a really good point. I'd like to keep it then for DX.

weaverryan:
This is awesome! Thank you Tito!

PR #110, opened 2021-02-10
Adding a simpler syntax for single stimulus controller elements

weaverryan:
Having 1 controller per element is the main use-case. This makes that a bit nicer to do :).

```twig
<div {{ stimulus_controller('chart', { 'name': 'Likes', 'data': [1, 2, 3, 4] }) }}>
    Hello
</div>
```

Also, this makes working with controllers that don't have any values (yet) MUCH nicer:

```twig
<!-- before -->
{{ stimulus_controller({ 'chart': {} }) }}

<!-- now -->
{{ stimulus_controller('chart') }}
```

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ See [stimulus-bridge](https://github.com/symfony/stimulus-bridge) for more detai
 For example:
 
 ```twig
-<div {{ stimulus_controller({ 'chart': { 'name': 'Likes', 'data': [1, 2, 3, 4] } }) }}>
+<div {{ stimulus_controller('chart', { 'name': 'Likes', 'data': [1, 2, 3, 4] }) }}>
     Hello
 </div>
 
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
If the goal is to make it easy for no-data controllers, shouldn't we remove the second argument and keep only one, string|array?
The polymorphism is not super nice (see the name of arguments...)

weaverryan:
> I'd the goal is to make it easy for no-data controllers, shouldn't we remove the second argument and keep only one, string|array?

Yes, that's the worst part, but why make 99% of the use-cases need an extra wrapping `{}`. The number of characters is not much different, but it's already hard to track all the `}` chars:

```
<div {{ stimulus_controller({ 'chart':  { 'name': 'Likes', 'data': [1, 2, 3, 4] } }) }}>

vs

<div {{ stimulus_controller('chart', { 'name': 'Likes', 'data': [1, 2, 3, 4] }) }}>
```

> The polymorphism is not super nice (see the name of arguments...)

VERY true. However, it's hidden (it's our burden) - and I don't see risk in doing this.

nicolas-grekas (approved):

tgalopin (approved):

PR #111, opened 2021-02-11
fix: fix EntrypointLookup Exception

jeremyFreeAgent:
Before it was:

```
Could not find the entry "fake_entry" in "...". Found: entrypoints.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,9 @@ matrix:
         - php: 7.1
         - php: 7.2
         - php: 7.3
-          env: deps=low
+          env:
+            - deps=low
+            - SYMFONY_PHPUNIT_VERSION=8.5
         - php: 7.4
           env: SYMFONY_PHPUNIT_VERSION=9.4
         - php: 8.0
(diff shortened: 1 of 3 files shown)


weaverryan:
Thanks for this @jeremyFreeAgent!

We're getting a test failure with the lowest deps... `expectExceptionMessageMatches` is not defined, which is odd since that happens on PHPUnit `7.5.20`, but that same version of PHP passes in another build on the matrix.

jeremyFreeAgent:
I didn't test it with that version in local. I'll look for it today to see why it is failing!

jeremyFreeAgent:
Work well with 7.3.26 and PHPUnit 7.5.20 in local. Very odd.

jeremyFreeAgent:
I reproduced the anomaly with installing dependencies as it is done in Travis (for 7.3 with low deps).
Maybe with `symfony/phpunit-bridge (5.x-dev)`.

weaverryan:
Thanks for checking into this further! Could we bump phpunit-bridge to a higher version? I don't see any downside to that, as it's a dev-dependency anyways.

jeremyFreeAgent:
I'm going to try that!

jeremyFreeAgent:
@weaverryan what do you think?

weaverryan:
Thanks Jérémy! Sorry for letting this sit for so long :)

jeremyFreeAgent:
Don’t worry! Happy to help ♥️

PR #112, opened 2021-02-16
Fix the safety of the stimulus_controller function

stof:
This function is safe in an html_attr context (and so automatically in the html context thanks to the way Twig treats html as a subset of html_attr). But the implementation does not ensure safety in other contexts.


weaverryan:
Thanks Stof!

Issue #103, opened 2021-01-25
defer always rendered for scripts in src/Asset/TagRenderer.php


lamselli:
In last version, I didn't want my scripts to have "defer" attribute, as I was using old jquery calls inside my rendered page.
But the defer attribute is added even when you set defer to false in config file (webpack_encore.yaml)
`script_attributes:
    defer: false`
    
As a workaround, I had to create a subscriber event to RenderAssetTagEvent to remove the attribute defer if its value is equal to false.

Sorry if I am unclear, but english is not my native language.

maxhelias:
Maybe an `array_filter` with a condition on the null value here : https://github.com/symfony/webpack-encore-bundle/blob/main/src/Asset/TagRenderer.php#L183, could be the solution.

weaverryan:
Fixed in #113. But also, even before that version, you can remove the `defer` attribute from the config :)

```diff
webpack_encore:
    script_attributes:
-        defer: true
```

Cheers!

PR #113, opened 2021-02-16
Fixing null and false attributes 

weaverryan:
Hi!

Suppose this config:

```yaml
webpack_encore:
    script_attributes:
        defer: false
        async: null
```

This PR changes what this outputs:

```html
<!-- before -->
<script src="..." defer="" async=""></script>

<!-- after -->
<script src="..." async></script>
```

This was an oversight when I built the feature.

With this PR, `false` correctly means "do not show this attribute". `null` is a bit trickier: I chose to make this mean "show the attribute, but with value (e.g. `<script defer>`). Any other values are rendered normally.

Cheers!

### DIFF
--- a/src/Asset/TagRenderer.php
+++ b/src/Asset/TagRenderer.php
@@ -177,12 +177,18 @@ class TagRenderer implements ResetInterface
 
     private function convertArrayToAttributes(array $attributesMap): string
     {
+        // remove attributes set specifically to false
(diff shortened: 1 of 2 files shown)


jmsche (approved):

maxhelias (approved):

maxhelias:
It fix this issue : https://github.com/symfony/webpack-encore-bundle/issues/103

Issue #94, opened 2020-07-15
Forward requests - Missing Assets after first forward.


Warxcell:
I found that https://github.com/symfony/webpack-encore-bundle/blob/master/src/Asset/EntrypointLookup.php#L84-L87
causes 1+N request to don't have assets.

This is my use case:

I have several private page that I need to export as PDF.
Because they are private (needs authentication) - I make several forward request to get generated HTML.
And then pass the generated HTML to service that generates PDF from HTML.

But all pages after first one have broken styles.

hktr92:
same issue happens if you try to render an HTML response in `kernel.exception`. after a few minutes of debugging I've came to the same conclusion as you.

~~as for a solution, I might come up with one after some further experiments.~~

I had to inject `EntrypointLookupInterface` service in my responder and call `reset()` method in order to return the `$newFiles` array. I've seen in #90 to discuss already about this one.

Warxcell:
currently I have this, which resolves the issue:

```
<?php

declare(strict_types=1);

namespace App\EventListener;

use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\HttpKernel\Event\ResponseEvent;
use Symfony\Component\HttpKernel\KernelEvents;
use Symfony\WebpackEncoreBundle\Asset\EntrypointLookupInterface;

class ResetAssets implements EventSubscriberInterface
{
    private EntrypointLookupInterface $entryLookup;

    public function __construct(EntrypointLookupInterface $entryLookup)
    {
        $this->entryLookup = $entryLookup;
    }

    public static function getSubscribedEvents()
    {
        return [
            KernelEvents::RESPONSE => 'resetAssets',
        ];
    }

    public function resetAssets(ResponseEvent $event)
    {
        $this->entryLookup->reset();
    }
}
```

But it's better to be fixed in vendor.

maxhelias:
Should be fixed no ?

PR #115, opened 2021-03-19
Reset assets on FINISH_REQUEST.

### LINKED ISSUE #90: Ability to "reset" cache of rendered files automatically in Twig

Warxcell:
Fixes #90, #94

### DIFF
--- a/src/DependencyInjection/WebpackEncoreExtension.php
+++ b/src/DependencyInjection/WebpackEncoreExtension.php
@@ -20,6 +20,7 @@ use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\WebLink\EventListener\AddLinkHeaderListener;
 use Symfony\WebpackEncoreBundle\Asset\EntrypointLookup;
 use Symfony\WebpackEncoreBundle\Asset\EntrypointLookupInterface;
+use Symfony\WebpackEncoreBundle\EventListener\ResetAssetsEventListener;
 
 final class WebpackEncoreExtension extends Extension
 {
@@ -57,6 +58,9 @@ final class WebpackEncoreExtension extends Extension
 
         $container->getDefinition('webpack_encore.entrypoint_lookup_collection')
             ->replaceArgument(0, ServiceLocatorTagPass::register($container, $factories));
+
+        $container->getDefinition(ResetAssetsEventListener::class)
+            ->setArgument(1, array_keys($factories));
         if (false !== $config['output_path']) {
             $container->setAlias(EntrypointLookupInterface::class, new Alias($this->getEntrypointServiceId('_default')));
         }
--- a/src/EventListener/ResetAssetsEventListener.php
+++ b/src/EventListener/ResetAssetsEventListener.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Symfony\WebpackEncoreBundle\EventListener;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\KernelEvents;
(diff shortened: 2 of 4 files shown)


weaverryan:
Sorry for the slow merge - thank you @Warxcell!

blankse:
@weaverryan After upgrading to v1.14.0 I have following error:
I have 2 entrypoints in my webpage. One on the top and one on the bottom. In v1.13.2 the runtime.js from webpack is only loaded in the entrypoint on the top. Now it is also loaded at the bottom and scripts are executed twice which makes makes mistackes (double click events...)

What can I do to have the runtime.js only on top?

Warxcell:
Hm, do you have sub-requests in-between? Can you provide reproducible code in some repo so I can fork and add a fix? Or directly add test case ?

blankse:
@Warxcell Yes there are sub-requests in-between. I can try to give you an reproducible code example.

noahduncan:
We're seeing similar issues with duplicate runtime.js script tags for our code.  I see use cases both ways, sometimes you want your all script tags on every request, sometimes you don't.  To me, that makes it feel like the option should be configurable  somewhere.

scrummer:
@Warxcell @weaverryan We also ran into the problem, that one of our projects kept loading the runtime.js multiple times. In our case it lead to forms being sent multiple times (including submission emails). Really had to look deep to figure out why it did that.
However; I kind of get the point why you guys implemented that feature. For us, this problem occured in the Pimcore ecosystem which uses (depending on the use-case) some up to a lot sub-requests.

As already stated; from a developers perspective I totally understand why you implemented that, but I think this sub-request thingy could break a lot of stuff. We were happy that it was "just" a contact form. I was wondering what would've happened if we updated the bundle in one of our larger shop systems (multiple orders? multiple items in the cart by accident? etc etc).

Unfortunately I can't really think of a "better" alternative than the Request-Finished event. But I think it really makes sense to think about it, how it could impact users.

Thank you, stay healthy!

BR,
Jan

H4wKs:
@Warxcell as @zing-james explained in #166, if you use the twig render function you end up with runtime.js being load twice, this is a huge side effect as I believe many people (including myself) use this in their projects.

kuldippujara:
Hi @Warxcell, @weaverryan,

I am having the same issue with embed controller after the update. I have check the other use cases where users want to reset the assets. I want to provide a suggestion that if we can differentiate the sub requests then we can put the condition when we want to reset and when we don't want to. 

1 option is to add attribute in request similar to _format eg. _renderer : inline / fragment
which will be only available if its render by Fragment renderer. 
then we can fix both the scenarios. 

waiting for feedback. 

Happy to create PR for Symfony and add attribute and then create PR to fix the issue.

Warxcell:
> 1 option is to add attribute in request similar to _format eg. _renderer : inline / fragment which will be only available if its render by Fragment renderer. then we can fix both the scenarios.


Sounds ok to me.

PR #121, opened 2021-04-23
[stimulus-controller] fix bool attributes from being rendered incorrectly

### LINKED ISSUE #118: stimulus_controller() improperly adds a value attribute when set to false
Hi!

Suppose you have this controller:

```js
// assets/controllers/testing_controller.js
import { Controller } from 'stimulus';

export default class extends Controller {
    static values = {
        isEnabled: Boolean
    }

    connect() {
        console.log(this.isEnabledValue);
    }
}
```

Then you enable this in a template:

```twig
<div
    {{ stimulus_controller('testing', {
        isEnabled: false
    }) }}
></div>
```

The expected behavior is that, on load, the log would say `false`. But, instead, you will get true. The cause is that the HTML is rendering like this:

```html
<div data-controller="testing" data-testing-is-enabled-value></div>
```

Notice the presence of the attribute (though it's not set to a value). It appears that, for boolean values, (we should triple check this) the presence of the attribute is enough to make it be set to `true`.

The fix would be to:

When a value is specifically set to `false`, we would render NO attribute. If it's set to any other value - even `null` - I think we DO need to set the attribute... because we don't know if the value is a Boolean type from PHP... or maybe a string. Though, I'm not sure if this really matters. But not setting the attribute for `null` or `''` seems... like we might be being too hands on, and the only issue I'm aware of currently is with the boolean.

Cheers!

jrushlow:
fixes #118

### DIFF
--- a/tests/IntegrationTest.php
+++ b/tests/IntegrationTest.php
@@ -214,7 +214,7 @@ class IntegrationTest extends TestCase
                 ],
             ],
             'controllerValues' => [],
-            'expected' => 'data-controller="my-controller" data-my-controller-boolean-value="1" data-my-controller-number-value="4" data-my-controller-string-value="str"',
+            'expected' => 'data-controller="my-controller" data-my-controller-boolean-value="true" data-my-controller-number-value="4" data-my-controller-string-value="str"',
         ];
 
         yield 'single-controller-nested-data' => [
@@ -247,7 +247,7 @@ class IntegrationTest extends TestCase
                 ],
             ],
             'controllerValues' => [],
-            'expected' => 'data-controller="symfony--ux-dropzone--dropzone" data-symfony--ux-dropzone--dropzone-my-key-value="1"',
+            'expected' => 'data-controller="symfony--ux-dropzone--dropzone" data-symfony--ux-dropzone--dropzone-my-key-value="true"',
         ];
 
         yield 'short-single-controller-no-data' => [
@@ -260,6 +260,18 @@ class IntegrationTest extends TestCase
             'dataOrControllerName' => 'my-controller',
             'controllerValues' => ['myValue' => 'scalar-value'],
             'expected' => 'data-controller="my-controller" data-my-controller-my-value-value="scalar-value"',
+        ];
+
+        yield 'false-attribute-value-renders-false' => [
+            'dataOrControllerName' => 'false-controller',
+            'controllerValues' => ['isEnabled' => false],
+            'expected' => 'data-controller="false-controller" data-false-controller-is-enabled-value="false"',
+        ];
+
+        yield 'true-attribute-value-renders-true' => [
+            'dataOrControllerName' => 'true-controller',
+            'controllerValues' => ['isEnabled' => true],
+            'expected' => 'data-controller="true-controller" data-true-controller-is-enabled-value="true"',
         ];
     }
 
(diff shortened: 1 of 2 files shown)


jrushlow:
related tests pass - failures related to deprecation that is fixed in #122

tests/IntegrationTest.php
```diff
@@ -261,6 +261,12 @@ public function provideRenderStimulusController()
             'controllerValues' => ['myValue' => 'scalar-value'],
             'expected' => 'data-controller="my-controller" data-my-controller-my-value-value="scalar-value"',
         ];
+
+        yield 'false-attribute-value-is-not-rendered' => [
+            'dataOrControllerName' => 'false-controller',
+            'controllerValues' => ['isEnabled' => false],
+            'expected' => 'data-controller="false-controller"',
+        ];
```
weaverryan, line 275: I did some more digging into this... and I think my proposed solution (which you implemented perfectly) is not... perfect ;).

For this test case:

```html
<!-- current output -->
<div
    data-controller="false-controller"
    data-false-controller-is-enabled-value=""
>
```

The `=""` for a `false` value is the problem. It should be `="false"`, which I've determined DOES cause the value in the controller to be `false`. Similarly, for `true`, the attribute should be `="true"`. Additionally, omitting the attribute entirely would cause `this.hasIsEnabledValue` (you can use a "hasser" like this to check for the existence of a value) to incorrectly return false with this current PR.

So I think the fix is to "fix" how we convert boolean values:

```
<!-- if isEnabled is passed as FALSE -->
<div
    data-controller="false-controller"
    data-false-controller-is-enabled-value="false"
>

<!-- if isEnabled is passed as TRUE -->
<div
    data-controller="false-controller"
    data-false-controller-is-enabled-value="true"
>
```

jrushlow: that make sense.. done. keeping this "unresolved" for future reference.

tests/IntegrationTest.php
```diff
@@ -214,7 +214,7 @@ public function provideRenderStimulusController()
                 ],
             ],
             'controllerValues' => [],
-            'expected' => 'data-controller="my-controller" data-my-controller-boolean-value="1" data-my-controller-number-value="4" data-my-controller-string-value="str"',
+            'expected' => 'data-controller="my-controller" data-my-controller-boolean-value="true" data-my-controller-number-value="4" data-my-controller-string-value="str"',
```
jrushlow, line 217: i've implemented `(string) true|false` for the rendered value. But this "breaks" the previous `1` === `true` output even though false was blank. 

I personally prefer `true|false` over `1|0` as it's clear that the value is not intended to be an `int` / `(string)` number. But would this be considered BC even though the PR is intended to fix `false` bool's?

hvt: Do not think this is a BC break, because one who defined a Stimulus value as `Boolean`, and the value was true-thy, having it previously rendered as `..-value="1"`, which in Stimulus was correctly interpreted as `true`.

And fixing that the `false` value was not being rendered as `..-value="false"` means imo it's currently incorrect, and of course the exact purpose of your PR ;].

hvt:
Just hit this in my project, and wanted to let you know this PR fixed it for me :].

jrushlow:
> Just hit this in my project, and wanted to let you know this PR fixed it for me :].

Awesome! That's good to hear - we should be able to get this pushed out pretty quickly.

weaverryan:
Thanks for the nice fix Jesse - and to @hvt also for double-checking it for us in a real app <3

PR #122, opened 2021-04-23
handle request deprecations

jrushlow:
- adds return type declarations where possible
- handles deprecated `HttpEvent::getMasterRequest()`


weaverryan:
Thanks Jesse!

Issue #123, opened 2021-04-27
Add classes support to {{ stimulus_controller }}


sukei:
The `stimulus_controller` function is a great addition but is not feature complete. Targets and actions are already discussed [here](https://github.com/symfony/webpack-encore-bundle/issues/119), so I won't talk about it. However classes are not covered. 

As stated by the [Stimulus documentation](https://stimulus.hotwire.dev/reference/css-classes):

> CSS class attributes must be specified on the same element as the data-controller attribute.

I think that `stimulus_controller` should add classes support too. But, I don't know how the function signature should change to support this without breaking things. Until the official implementation arrived, I came to the following:

`stimulus_controller('layout/panel', { active: true }, { hide: d-none' }, 'focusable')` which rendered `data-controller="layout--panel focusable" data-layout--panel-active-value="true" data-layout--panel-hide-class="d-none"`.

Kocal:
The problem is how it should behave when using the array syntax to render multiple controllers values like this?
```twig
{{ stimulus_controller({
  'controller1': { 'value1' => '...', 'value2': '...' },
  'controller2': { 'value1' => '...', 'value2': '...' },
}) }}
```

Maybe a 2nd parameter like this:
```twig
{{ stimulus_controller({
  'controller1': { 'value1' => '...', 'value2': '...' },
  'controller2': { 'value1' => '...', 'value2': '...' },
}, {
  'controller1': {'loading': 'my-class-for-loading'}
}) }}
```

WDYT?

If it's not readable enough and/or hard to maintain, maybe we can add a new argument and exploit [Twig named arguments feature](https://twig.symfony.com/doc/3.x/templates.html#named-arguments) ?
```twig
{{ stimulus_controller({
  'controller1': { 'value1' => '...', 'value2': '...' },
  'controller2': { 'value1' => '...', 'value2': '...' },
}, classes={
  'controller1': {'loading': 'my-class-for-loading'}
}) }}

sukei:
The current API may be hard to keep. That's the main issue here. My own version uses a triplet of values to achieve this purpose (`controller, values, classes, controller, values, classes, ...`). However, the current API can be tweaked to get the same signature. Currently we have

`$dataOrControllerName, array $controllerValues = []`

Could change to:

`string|array $controller /* array of (string, mixed) to keep BC */, array $controllerValues = [], array $controllerClasses = [], ...$repeat`

weaverryan:
[...]

PR #124, opened 2021-04-27
feat(twig): implements stimulus_action() and stimulus_target() Twig functions, close #119

### LINKED ISSUE #119: Add {{ stimulus_action }} / {{ stimulus_target }}
``{{ stimulus_controller() }}`` is a great way to assign values to controller but targets and actions requires fallback to long syntax.

Can we have something like this:

```twig
<div {{ stimulus_controller('pages/reports/table') }}>

    <a {{ stimulus_action('pages/reports/table', 'onClick') }}>Click me</a>
    <span {{ stimulus_target('pages/reports/table', 'results') }}>results here</span>

</div>
```

Otherwise, we have to write something like this:

```twig

<div {{ stimulus_controller('pages/reports/table') }}>

    <a data-action="pages--reports--table#onClick">Click me</a>
    <span data-pages--reports--table-target="results">results here</span>

</div>
```

which is harder to write. Having those 2 extensions would also allow easier moving/renaming controllers; users would only need to find ``'pages/reports/table'`` and replace with ``'pages/zones/table-advanced'``.

---

``{{ stimulus_action('pages/reports/table', 'onClick') }}`` would also allow third parameter when non-default event is used:

```twig
<a {{ stimulus_action('pages/reports/table', 'onNonDefaultEvent', 'hover') }}>Hover me</a>
```
 
which would generate

```twig
<a data-action="hover->pages--reports--table#onNonDefaultEvent">Hover me</a>
```

Kocal:
Hi :wave: 

This PR is a proposal for #119, which adds `stimulus_action()` and `stimulus_target()` Twig functions.

`stimulus_action()` can be a bit hard to use because it supports many syntaxes, since Stimulus allows to use multiple actions in a single `data-action` attribute and we should takes care of the default event:
- `stimulus_action('controller', 'method')` 
- a custom event: `stimulus_action('controller', 'method', 'event)`
- multiple controllers/actions: 
```twig
{{ stimulus_action({
  'controller': 'method',
  'controller-2': ['method', 'method2'],
  'controller-3': {'click': 'onClick'},
  'controller-4': ['method', {'click': 'onClick'}, {'window@resize': 'onWindowResize'}],
}) }}
```

For `stimulus_target`:
- `stimulus_target('controller', 'target')`
- `stimulus_target('controller', 'target1 target2')`, Stimulus allows multiple targets
- `stimulus_target({ 'controller1': 'target', 'controller2': 'target2 target3' })`, array syntax

Usage:
```twig
<div {{ stimulus_controller('foo') }}>
  <div {{ stimulus_target('foo', 'myTarget') }}></div>
  <!-- <div data-foo-target="myTarget"></div> -->

  <div {{ stimulus_target('foo', 'myTarget mySecondTarget') }}></div>
  <!-- <div data-foo-target="myTarget mySecondTarget"></div> -->

  <div {{ stimulus_target({ 'foo': 'myTarget', 'bar': 'anotherTarget' }) }}></div>
  <!-- <div data-foo-target="myTarget" data-bar-target="anotherTarget"></div> -->

  {# Listen to default event and call "onClick" #}
  <a {{ stimulus_action('foo', 'onClick') }}>A link</a>
  <!-- <div data-action="foo#onClick">A link</a> -->

  {# Listen to event "click" and call "onClick" #}
  <a {{ stimulus_action('foo', 'onClick', 'click') }}>A Link</a>
  <!-- <div data-action="click->foo#onClick">A link</a> -->

  {# Listen to default event, call "onClick" from "foo" controller and "onClick" + "onSomethingElse" from "bar" controller #}
  <a {{ stimulus_action({
    'foo': 'onClick', 
    'bar': ['onClick', 'onSomethingElse']
  }) }}>A link</a>
  <!-- <div data-action="click->foo#onClick bar#onClick bar#onSomethingElse">A div</div> -->

  {# It is possible to pass a map, here "bar#onWindowResize" will be called on "window@resize" #}
  <div {{ stimulus_action({
    'foo': 'onClick', 
    'bar': ['onClick', {'window@resize': 'onWindowResize'}]
  }) }}>A div</a>
  <!-- <div data-action="foo#onClick bar#onClick window@resize->bar#onWindowResize">A div</div> -->
</div>
```

WDYT? Thanks!

---

For https://github.com/symfony/webpack-encore-bundle/issues/119#issuecomment-827356762, I think it should belongs to another PR.

### DIFF
--- a/src/Twig/StimulusTwigExtension.php
+++ b/src/Twig/StimulusTwigExtension.php
@@ -19,6 +19,8 @@ final class StimulusTwigExtension extends AbstractExtension
     {
         return [
             new TwigFunction('stimulus_controller', [$this, 'renderStimulusController'], ['needs_environment' => true, 'is_safe' => ['html_attr']]),
+            new TwigFunction('stimulus_action', [$this, 'renderStimulusAction'], ['needs_environment' => true, 'is_safe' => ['html_attr']]),
+            new TwigFunction('stimulus_target', [$this, 'renderStimulusTarget'], ['needs_environment' => true, 'is_safe' => ['html_attr']]),
         ];
     }
 
@@ -27,7 +29,7 @@ final class StimulusTwigExtension extends AbstractExtension
      *                                           as keys set to their "values". Or this
      *                                           can be a string controller name and data
      *                                           is passed as the 2nd argument.
-     * @param array $controllerValues Array of data if a string is passed to the first argument.
+     * @param array $controllerValues Array of data if a string is passed to the 1st argument.
      * @return string
      * @throws \Twig\Error\RuntimeError
      */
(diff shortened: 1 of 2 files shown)


src/Twig/StimulusTwigExtension.php
```diff
@@ -73,6 +73,79 @@ public function renderStimulusController(Environment $env, $dataOrControllerName
...
+     *                                           as keys set to their "actions" and "events".
+     *                                           Or this can be a string controller name and
+     *                                           action and event are passed as the 2nd and 3rd arguments.
+     * @param string|null  $actionName The action to trigger if a string is passed to the 1st argument. Optional.
+     * @param string|null  $eventName  The event to listen to trigger if a string is passed to the 1st argument. Optional.
+     *
+     * @return string
+     * @throws \Twig\Error\RuntimeError
+     */
+    public function renderStimulusAction(Environment $env, $dataOrControllerName, string $actionName = null, string $eventName = null): string
```
moismailzai, line 89: Given that we're already documenting it as such in the doc block, does it make sense to use a union type for `$dataOrControllerName`? This would apply to the `renderStimulusAction` method as well.
```suggestion
    public function renderStimulusAction(Environment $env, string|array $dataOrControllerName, string $actionName = null, string $eventName = null): string
```

src/Twig/StimulusTwigExtension.php
```diff
@@ -73,6 +73,79 @@ public function renderStimulusController(Environment $env, $dataOrControllerName
...
+                }
+            }
+        }
+
+        return rtrim('data-action="'.implode(' ', $actions).'"');
+    }
+
+    /**
+     * @param string $controllerName The controller name.
+     * @param string $targetName The target name.
```
moismailzai: According to the [docs](https://stimulus.hotwire.dev/reference/targets#attributes-and-names), it's valid to provide multiple space-separated targets. Would it be useful to mention that in the doc block?

```suggestion
     * @param string $targetNames The space-separated list of target names.
```

src/Twig/StimulusTwigExtension.php
```diff
@@ -73,6 +73,79 @@ public function renderStimulusController(Environment $env, $dataOrControllerName
...
+     *
+     * @return string
+     * @throws \Twig\Error\RuntimeError
+     */
+    public function renderStimulusTarget(Environment $env, string $controllerName, string $targetName): string
```
moismailzai: Given that the other two helper functions allow us to pass in an array, would it be a more consistent user-experience if this functioned in the same way?

moismailzai (commented):
This looks good to me, just a few thoughts!

src/Twig/StimulusTwigExtension.php
Kocal: 👎 for me, this syntax is only compatible from PHP 8 (where the bundle also requires to be compatible with PHP 7.1.3+)

src/Twig/StimulusTwigExtension.php
Kocal: Oh I wasn't aware about this syntax, thanks!

src/Twig/StimulusTwigExtension.php
Kocal: Do you mean something like this?
```twig
{{ stimulus_target({
  'controller': 'target1 target2',
  'controller-2': 'target3 target4',
}) }}
```

src/Twig/StimulusTwigExtension.php
moismailzai: that's a very good point, agreed!

src/Twig/StimulusTwigExtension.php
moismailzai: Yes, exactly, that would be consistent with the other helpers.

Kocal: Added in 446be773b5c1f0c903e9fc4c01d221dc99446d17

moismailzai (approved):
Overall the proposed syntax / documentation / implementation looks good to me. Seeing some duplicated logic in the methods but I think there may be some more work done in here (eg. https://github.com/symfony/webpack-encore-bundle/issues/123) so maybe not a good idea to start consolidating logic yet. +1 from me.

src/Twig/StimulusTwigExtension.php
```diff
@@ -73,6 +73,103 @@ public function renderStimulusController(Environment $env, $dataOrControllerName
...
+     *                                           as keys set to their "actions" and "events".
+     *                                           Or this can be a string controller name and
+     *                                           action and event are passed as the 2nd and 3rd arguments.
+     * @param string|null  $actionName The action to trigger if a string is passed to the 1st argument. Optional.
+     * @param string|null  $eventName  The event to listen to trigger if a string is passed to the 1st argument. Optional.
+     *
+     * @return string
+     * @throws \Twig\Error\RuntimeError
+     */
+    public function renderStimulusAction(Environment $env, $dataOrControllerName, string $actionName = null, string $eventName = null): string
```
weaverryan, line 89: Don't forget to expose these new functions up in `getFunctions()` ;)

weaverryan (commented):
I like this a lot! The super advanced syntaxes are ugly, but logical - and they will be rarely used. The simple syntax is... nice and simple :). Good work!

src/Twig/StimulusTwigExtension.php
Kocal: OMG yes :scream: 

I totally forgot about them, usually I write tests for Twig Extensions by creating a fake loader and doing assertions on `$twig->render()` This way I can test the extension and its integration with Twig.

Thanks! :smile:

maxhelias (approved):

Kocal:
Friendly ping @weaverryan @tgalopin 

Thanks!

weaverryan:
As I don't see any other feedback, I'm going to merge this. Thanks @Kocal! But... would you mind opening a PR to the README to document these? I'd like to have that before we release :)

Kocal:
Thanks @weaverryan, I'm on it!

Kocal:
Opened at #125

PR #126, opened 2021-05-20
Remove fallback cache on cache warmer

deguif:
Since Symfony 4.2 fallback cache on cache warmer is not supported anymore.

Here's the PR that dropped it: https://github.com/symfony/symfony/pull/28331

### DIFF
--- a/src/CacheWarmer/EntrypointCacheWarmer.php
+++ b/src/CacheWarmer/EntrypointCacheWarmer.php
@@ -9,7 +9,6 @@
 
 namespace Symfony\WebpackEncoreBundle\CacheWarmer;
 
-use Psr\Cache\CacheItemPoolInterface;
 use Symfony\Bundle\FrameworkBundle\CacheWarmer\AbstractPhpFileCacheWarmer;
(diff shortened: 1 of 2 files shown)


src/CacheWarmer/EntrypointCacheWarmer.php
```diff
@@ -22,7 +22,7 @@ class EntrypointCacheWarmer extends AbstractPhpFileCacheWarmer
     public function __construct(array $cacheKeys, string $phpArrayFile, CacheItemPoolInterface $fallbackPool)
```
deguif: Not sure if we can drop the `$fallbackPool` as it would be a BC break

weaverryan: I think it's ok to remove it

deguif: Ok I'll do it then ;)

deguif: Done

weaverryan:
Thanks @deguif!

PR #128, opened 2021-05-24
Add return type-hints to twig extensions

deguif:
This should not be a BC break as classes are marked as final


weaverryan:
Thanks again @deguif!

PR #130, opened 2021-06-02
Fix error message in EntrypointLookup

tgalopin:
Currently, the only listed entrypoint is "entrypoints"


weaverryan:
Duplicate of #111 - but thanks for the poke that I need to do some merging and tagging :)

PR #136, opened 2021-07-18
Allow Symfony 6

Kocal:
Symfony 6 is not released yet, but I suppose it can be a good idea to starts to add Symfony 6 support to bundles.﻿

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -59,7 +59,7 @@ jobs:
         uses: ramsey/composer-install@v1
 
       - name: Psalm
-        run: psalm --no-progress --output-format=github
+        run: psalm --no-progress --php-version=8.0
 
   composer-normalize:
     name: Composer Normalize
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -53,7 +53,7 @@ jobs:
         with:
           php-version: 8.0
           coverage: none
-          tools: vimeo/psalm:4.8.1
+          tools: vimeo/psalm:4.11.2
 
       - name: Download dependencies
         uses: ramsey/composer-install@v1
--- a/composer.json
+++ b/composer.json
@@ -11,17 +11,17 @@
     ],
     "require": {
         "php": ">=7.1.3",
-        "symfony/asset": "^4.4 || ^5.0",
-        "symfony/config": "^4.4 || ^5.0",
-        "symfony/dependency-injection": "^4.4 || ^5.0",
-        "symfony/http-kernel": "^4.4 || ^5.0",
+        "symfony/asset": "^4.4 || ^5.0 || ^6.0",
+        "symfony/config": "^4.4 || ^5.0 || ^6.0",
+        "symfony/dependency-injection": "^4.4 || ^5.0 || ^6.0",
+        "symfony/http-kernel": "^4.4 || ^5.0 || ^6.0",
         "symfony/service-contracts": "^1.0 || ^2.0"
     },
     "require-dev": {
-        "symfony/framework-bundle": "^4.4 || ^5.0",
-        "symfony/phpunit-bridge": "^5.3",
-        "symfony/twig-bundle": "^4.4 || ^5.0",
-        "symfony/web-link": "^4.4 || ^5.0"
+        "symfony/framework-bundle": "^4.4 || ^5.0 || ^6.0",
+        "symfony/phpunit-bridge": "^5.3 || ^6.0",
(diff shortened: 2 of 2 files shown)


jrushlow:
Howdy @Kocal - I do not believe we're ready to open up the bundle's to Symfony 6 yet based on based internal discussions. Symfony 5.4 has not been released yet and Symfony 6 is still far to unstable.

Going to close for now - we can reopen / merge in the future when 5.4 has been released and 6 is a bit more stable.
cc @weaverryan

Kocal:
Hi!

The [Twig Extra Bundle](https://github.com/twigphp/twig-extra-bundle/commit/6a4494b3c4df73bd170cf1f472ac7930eb9949ae) already supports Symfony 6 since 3 months, that's why I started PRs to allow Symfony 6 on many bundles, the earlier the better (to avoid the frustration people had when upgrading to a new Symfony major version, but no bundles supports it), but I suppose this is an exception.

javiereguiluz:
I agree with @Kocal and please note that Symfony 6.0 will end its development in only 11 weeks (and then it's start an 8-week stabilization phase). See https://symfony.com/releases

@jrushlow but of course you are in charge of managing this bundle and you know better how to handle things, so we don't want to stress you about this 🙏  Thanks!

Kocal:
Re-opened: https://twitter.com/nicolasgrekas/status/1438831963838816262?s=19 😌

jrushlow:
@Kocal can you rebase this? Thanks in advance!

Kocal:
I don't have access to a computer until Monday in the morning, so I can't rebase right now :/

Kocal:
@jrushlow PR rebased.

The Psalm job is failing, unlucky it does not output anything that can help us in the CI. This is what it outputs in local:
image

Should I add those errors in the psalm baseline? Thanks

jrushlow:
Thanks @Kocal - yes, adding those to the baseline for now is probably best. The problem ultimately is in Psalm itself as `mixed` is not a class but a PHP type in the referenced instances. Either way that has nothing to do with this PR so we should be able to get this merged and released pretty quick.

Kocal:
@jrushlow the PR has been rebased.

How can I generate Psalm baseline? I've tried to run it with `--update-baseline` but it does not ignore the 3 errors :(

williarin:
https://github.com/symfony/webpack-encore-bundle/blob/5a86aad75424096b47808423f27da87d8bebab8e/.github/workflows/static.yml#L56

Update psalm to latest version 4.11.2 and the test passes.

Kocal:
Updating to psalm 4.11.2 didn't work. 

How can I generate the baseline _again_? With PHPStan you can do it very easily with `--generate-baseline` but with Psalm?

Thanks!

GwendolenLynch:
> How can I generate the baseline again? With PHPStan you can do it very easily with --generate-baseline but with Psalm?

`psalm --update-baseline` but that doesn't seem to fix the issue either. 

After a quick bit of debugging what I can see is that Psalm's `MethodCallReturnTypeFetcher#fetch` is assuming that the return type of `\Symfony\Component\HttpFoundation\ParameterBag#get` is a class name and therefore matching `int|float|bool|string|void|null|false|true|object|mixed` will result in the error. 

Downgrading `symfony/http-foundation` to `5.4-dev` stops this from triggering as the return type is still defined as a PHPDoc, but given the whole point here is to support Symfony 6 :woman_shrugging:

weaverryan:
Hi everyone!

Thanks @Kocal for getting this going and continually updating it - and @GwendolenLynch for jumping in to help. I don't know the exact fix, but the problem is that Psalm is running in "PHP 7.1" mode. Check out this psalm config: https://psalm.dev/docs/running_psalm/configuration/#phpversion

Because this isn't specified, it's apparently falling back to the PHP version in composer.json, which is `>=7.1.3`. So I assume that it is using "php 7.1" mode. 

In theory... we should maybe run psalm twice - once with the lowest php version and once with the highest, but that feels overkill. I would be ok with setting the `phpVersion` config to `8.0`, preferably with a `--php-version=8.0` command line flag, so we don't forget that the version is hiding in a config file :).

Cheers!

jrushlow (approved):

xabbuh:
Would it make sense to add a new job to the CI that would actually run the tests against Symfony 6?

weaverryan:
Thank you @Kocal! I'm going to look at the CI matrix in a separate PR

Kocal:
Perfect! :)

PR #138, opened 2021-07-23
Allow Symfony 6

Nyholm:


nicolas-grekas:
~I think it's too early to allow Symfony 6 anywhere. The reason is that we're not done yet with adding BC breaks (aka return types) to it.~

Nyholm:
Allowing people (CI) to install this bundle with Symfony CI is not the same as guarantee that it will work. 

I understand what using dev-main means. I know that Im really living on the edge and things might/will change. 

Strictly speaking: We are not done adding BC breaks until November.

nicolas-grekas:
Then, merging this PR will block any releases of the bundle.

Nyholm:
No, I would not agree with that. =)

jdreesen:
This was also discussed in #136

Nyholm:
I slept on in. I see what @nicolas-grekas is saying. 

> I think it's too early to allow Symfony 6 anywhere. The reason is that we're not done yet with adding BC breaks (aka return types) to it.

You mean to THIS package. I though you referred to Symfony. 

I agree with you. Merging this PR now is premature.

nicolas-grekas:
Topic unlocked! I also reopened #136. I'll let you decide which one to keep and which one to close :)

jrushlow:
going to close this in favor of PR #136 as it was opened first. Thanks @Nyholm for the help on this!

PR #139, opened 2021-07-23
Use GitHub CI instead of Travis

Nyholm:
This CI may not perfect but it is better than not running it at all


Nyholm:
The CI is red, but it would be great with a merge and next PR can make sure things start to look green

.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
...
+          - php: '7.2'
+            sf_version: '*'
+          - php: '7.3'
+            sf_version: '*'
+          - php: '7.4'
+            sf_version: '*'
+
+    steps:
+      - name: Set up PHP
+        uses: shivammathur/setup-php@2.7.0
```
Kocal, line 31: What about using `shivammathur/setup-php@^2`? This way we wouldn't have to care about the action's version in the future for each bug fix/new features.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
...
+  pull_request: ~
+  push: ~
+
+jobs:
+  phpunit:
+    name: PHPUnit
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 10
+      matrix:
```
Kocal, line 14: IMO it would be nice to configure the strategy as `fail-fast: false` (see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast).
By doing this, if a job fails, the other jobs won't be stopped and will continue to run.

.github/workflows/static.yml
```diff
@@ -0,0 +1,80 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 8.0
+          coverage: none
+          tools: phpstan:0.12.92, cs2pr
```
Kocal, line 18: What about required PHPStan as dev dependencies? This way we can run it locally before pushing.

.github/workflows/static.yml
```diff
@@ -0,0 +1,80 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 8.0
+          coverage: none
+          tools: php-cs-fixer:3.0.0, cs2pr
```
Kocal, line 39: Same for PHP CS Fixer

Kocal (commented):
Great initiative, thanks!
I have some minor comments.

.github/workflows/ci.yml
Nyholm: I prefer to specify the version so nothing unexpected changes. Currently we don’t have to care about future bugs in setup-php. 

If we want to use a new feature, we can bump the version then.

.github/workflows/ci.yml
Nyholm: Sure. Good idea.

.github/workflows/static.yml
Nyholm: Hard no. :)

Static analyzers are not a dependency of this project. It is just a tool to help you develop. Just like PHPstorm is a tool that will help you develop. 

Experienced developers can run these tools (installed globally) before they push. If you don’t, you will get an automated PR comment of things you should change. 

We could add a Makefile script to download and run static analyzers. I think that is overkill, but anyways it is outside the scope of this PR.

Kocal: Alright, I just wanted your point of view about this, since I'm in the "install locally tooling dependencies" team.

Thanks! And nice job as usual :)

Nyholm: Yeah. I definitely see the value of your "team" when php-cs-fixer updated to a new major version or if your project is doing something psalm-specific. 

But I still find this to be the simplest, most automated way forward. Especially for small bundles/libraries like this one.

weaverryan:
Thanks Tobias!

phpstan-baseline.neon
```diff
@@ -0,0 +1,62 @@
+parameters:
```
weaverryan, line 1: pslam AND phpstan?

.github/workflows/.editorconfig
```diff
@@ -0,0 +1,2 @@
+[*.yml]
+indent_size = 2
```
weaverryan, line 2: Is this usually committed? If it's only to control tabs when modifying .yml files in this directory, it feels kinda overkill

phpstan-baseline.neon
Nyholm: Yes sir!

They actually complement each other and they are very easy to run.

.github/workflows/.editorconfig
Nyholm: Yes and no.. 

There should be an .editorconfig in root too. Then this one will make (more) sense.

PR #140, opened 2021-07-24
Update .gitattributes

Nyholm:
I'll keep making some chore PRs.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,10 @@
+/.github
 /tests export-ignore
 /phpunit.xml.dist export-ignore
 /.gitattributes export-ignore
 /.gitignore export-ignore
 /.php-cs-fixer.dist.php export-ignore
-/.travis.yml export-ignore
+/phpstan.neon.dist
+/phpstan-baseline.neon
+/psalm.baseline.xml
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,6 +1,10 @@
...
 /tests export-ignore
 /phpunit.xml.dist export-ignore
 /.gitattributes export-ignore
 /.gitignore export-ignore
 /.php-cs-fixer.dist.php export-ignore
-/.travis.yml export-ignore
+/phpstan.neon.dist
+/phpstan-baseline.neon
+/psalm.baseline.xml
+/psalm.xml
```
weaverryan, line 10: I noticed that none of these new items are consistent with what's done on symfony/symfony - e.g. https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Cache/.gitattributes - can you help me understand why these should be added here but not there?

Nyholm: The symfony/symfony package should not be used. I don't really care about that package (but yes, we should maybe add `psalm.xml` there. 

The cache component does not have any of these files.

weaverryan:
Thanks Tobias!

PR #147, opened 2021-10-15
Add encore_entry_exists() twig functions to check if entrypoint has files

### LINKED ISSUE #100: Allow to check if an entry exists for a given entrypoint
Hi, 
I'd like to display script tags only if the entry exists using a Twig function like `encore_entry_script_tags_if_exists`.
As I know, there is actually no public method to check if an entry is defined for a given entry point.

Do you think this is possible to add such a behavior?

Thanks

acrobat:
I've added 2 helper twig functions to check if the entrypoint exists and if it has files. 

Closes #100

### DIFF
--- a/src/Asset/EntrypointLookup.php
+++ b/src/Asset/EntrypointLookup.php
@@ -140,4 +140,11 @@ class EntrypointLookup implements EntrypointLookupInterface, IntegrityDataProvid
 
         return $this->entriesData;
     }
+
+    public function entryExists(string $entryName): bool
+    {
+        $entriesData = $this->getEntriesData();
+
+        return isset($entriesData['entrypoints'][$entryName]);
+    }
 }
--- a/src/Twig/EntryFilesTwigExtension.php
+++ b/src/Twig/EntryFilesTwigExtension.php
@@ -10,6 +10,7 @@
 namespace Symfony\WebpackEncoreBundle\Twig;
 
 use Psr\Container\ContainerInterface;
+use Symfony\WebpackEncoreBundle\Asset\EntrypointLookup;
 use Symfony\WebpackEncoreBundle\Asset\EntrypointLookupInterface;
 use Symfony\WebpackEncoreBundle\Asset\TagRenderer;
 use Twig\Extension\AbstractExtension;
@@ -31,6 +32,7 @@ final class EntryFilesTwigExtension extends AbstractExtension
             new TwigFunction('encore_entry_css_files', [$this, 'getWebpackCssFiles']),
             new TwigFunction('encore_entry_script_tags', [$this, 'renderWebpackScriptTags'], ['is_safe' => ['html']]),
             new TwigFunction('encore_entry_link_tags', [$this, 'renderWebpackLinkTags'], ['is_safe' => ['html']]),
+            new TwigFunction('encore_entry_exists', [$this, 'entryExists']),
         ];
     }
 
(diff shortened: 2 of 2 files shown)


weaverryan:
Hi!

Can you explain your use-case a bit more? If your entry doesn't have an CSS or JS files, they already won't be rendered. But perhaps you have some situation where an entrypoint may or may not be even defined (?) and you need to be able to have `encore_entry_script_tags('foo')` without generating an error that the entry doesn't exist?

Cheers!

acrobat:
Hi @weaverryan,

Yes indeed the last part is our use case. We are in the progress of replacing a custom build layer (on top of webpack) with symfony encore in [our Kunstmaan CMS](https://github.com/kunstmaan/KunstmaanBundlesCms). If you install the cms your project has 2 "setup's" an "app" entry (or even split in multiple smaller entries) for the frontend of the website and an "admin" entry which users can define (but are not required to) to override css/js of the admin interface.

The switch to symfony encore is almost ready, but we had indeed the issue that the call `encore_entry_script_tags('admin')` for the admin entry (which might not be defined in a user project) throws an exception. https://github.com/symfony/webpack-encore-bundle/blob/5a86aad75424096b47808423f27da87d8bebab8e/src/Asset/EntrypointLookup.php#L95-L102

So that's the reasoning behind this PR, but as you suggested not throwing or catching the exception in `encore_entry_script_tags` and `encore_entry_link_tags` would also fix our case. I don't know which approach you prefer, but I'm able to update the PR if needed!

src/Twig/EntryFilesTwigExtension.php
```diff
@@ -29,6 +30,8 @@ public function getFunctions(): array
         return [
             new TwigFunction('encore_entry_js_files', [$this, 'getWebpackJsFiles']),
             new TwigFunction('encore_entry_css_files', [$this, 'getWebpackCssFiles']),
+            new TwigFunction('encore_has_entry_js_files', [$this, 'hasWebpackJsFiles']),
+            new TwigFunction('encore_has_entry_css_files', [$this, 'hasWebpackCssFiles']),
```
weaverryan: What about a single function: encore_entry_exists()? Because it is not about whether or not there are css or js files in this entry - just whether the entry is defined at all, right?

acrobat: that would work too! I will adjust the PR

acrobat: I've updated the code!

Exellent1988:
is there any progress regaring the pull request? I'm also quite interested in this function to check if an tag has files / exists at all.

acrobat:
@weaverryan I think this PR is ready for a final review, let me know if there are any changes needed!

acrobat:
Friendly ping @weaverryan

src/Twig/EntryFilesTwigExtension.php
```diff
@@ -58,6 +60,19 @@ public function renderWebpackLinkTags(string $entryName, string $packageName = n
             ->renderWebpackLinkTags($entryName, $packageName, $entrypointName, $attributes);
     }
 
+    public function entryExists(string $entryName, string $entrypointName = '_default'): bool
+    {
+        try {
+            $entrypointLookup = $this->getEntrypointLookup($entrypointName);
+            $entrypointLookup->getJavaScriptFiles($entryName);
+            $entrypointLookup->reset();
```
weaverryan: Hmm. I don't love this part.

Instead, what if we add an `entryExists()` method to `EntrypointLookup`. It would NOT be part of the interface - but a bonus method on that class.

Then, in here, we would have something like:

```php
$entrypointLookup = $this->getEntrypointLookup($entrypointName);
if (!$entrypointLookup instanceof EntrypointLookup) {
    throw new \LogicException(sprintf('Cannot use entryExists() unless the entrypoint lookup is an instance of "%s", EntrypointLookup::class));
}

return $entrypointLookup->entryExists($entryName);
```

This will avoid the "reset()" part, which could, in theory, have some side effects (if you have, somehow, already rendered some files from this entrypointlookup).

acrobat: Sounds like a good solution to me, I've updated the PR with the requested changes!

src/Asset/EntrypointLookup.php
```diff
@@ -140,4 +140,9 @@ private function getEntriesData(): array
 
         return $this->entriesData;
     }
+
+    public function entryExists(string $entryName): bool
+    {
+        return isset($entriesData['entrypoints'][$entryName]);
```
jdreesen, line 148:
```suggestion
        return isset($this->entriesData['entrypoints'][$entryName]);
```

weaverryan:
Thanks @acrobat!

PR #152, opened 2021-11-24
README.md : Fix events order

Arkounay:
The correct way to call the global window resize event is `resize@window` and not `window@resize`


Kocal (approved):

Kocal:
Thanks!

weaverryan:
Thanks @Arkounay!

PR #153, opened 2021-11-25
Skipping null values from rendering

sadikoff:
Hi there!

This should fix a weird situation when we are rendering stringified `null` value attribute and stimulus using `null` string as a value instead of just skipping it.


weaverryan:
Thanks for this @sadikoff!

To be fully clear, if you passed a true `null` to `stimulus_controller()` for a value that is a `String` type in your Stimulus controller, then:

**Before this change**: The value would be the string `null` (literally those 4 charaters) in JavaScript.
**After this change**: The value will be `''` empty quotes. It still won't be a true `null` in your controller, but that's just because Stimulus expects the type to be a `String` always, so it gives you a astring.

PR #155, opened 2021-12-01
Increase version constraint of symfony/service-contracts

luca-rath:
This allows to install `symfony/webpack-encore-bundle` in a new Symfony 6 project without it being necessary to downgrade `psr/container` and `symfony/service-contracts`

Fixes #155


weaverryan:
Thank you for this Luca!

PR #167, opened 2022-03-20
Add `entryExists` to EntrypointLookupInterface

lcharette:
I don't know why it was proposed in [#147](https://github.com/symfony/webpack-encore-bundle/pull/147#discussion_r803759695) _not_ to include `entryExists` in `EntrypointLookupInterface`, but it doesn't makes sense to test agains't a specific class instead of the interface.

In my case, this cause an issue as I cannot use dependence injection on `EntrypointLookupInterface` and expect `entryExists` to be present. 

If this is not approved, I suggest adding `entryExists` to it's own interface and have `EntrypointLookup` implement both (although you won't fix the dependency injection problem).

### DIFF
--- a/src/Asset/EntrypointLookupInterface.php
+++ b/src/Asset/EntrypointLookupInterface.php
@@ -28,4 +28,6 @@ interface EntrypointLookupInterface extends ResetInterface
      * Resets the state of this service.
      */
     public function reset();
+
+    public function entryExists(string $entryName): bool;
 }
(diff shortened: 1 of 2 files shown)


weaverryan:
Sorry for the slow reply. Not breaking BC was the reason behind not adding it to the interface, which I agree is a bit awkward. We have two choices:

A) Add a 2nd interface as you suggested
B) Add the new method to the interface in a BC-friendly way (this includes documenting the new method on the interface, then we would actually make it a *real* method in the next major version - the `UserInterface::getUserIdentifier()` is an example of doing that - https://github.com/symfony/symfony/blob/1b57d282713423be349104fd2c9786c9eda2fc38/src/Symfony/Component/Security/Core/User/UserInterface.php#L29

So, we can't merge this PR - but I'd welcome a PR with either of these approaches.

Cheers!

PR #169, opened 2022-04-20
prevent reset assets on FINISH REQUEST for sub request type to fix th…

kuldippujara:
prevent reset assets on FINISH REQUEST for sub request type to fix the issue of embedding controller inside twig files.


weaverryan:
Thanks! I just merged #172 for this :)

PR #171, opened 2022-04-21
Do not JSON encode stringable values

jderusse:
When using `stimulus_controller` with values generated by twig, the values are converted to JSON. Which prevent the following code working out of the box.

```twig
{% set tooltip %}
  Hello <strong>{{ name }}</strong>
{% endset %}

<div {{ stimulus_controller('tooltip', {html: true, title: tooltip}) }}>
  title
</div>
```

Because the content of the variable tooltip is an `\Twig\Markup` object.

This PR stringifies `values` instead of json_encoding.


weaverryan:
That's a nice catch - thank you Jérémy!

PR #172, opened 2022-05-03
Fixing reset assets trigger on sub-requests

TarikAmine:
`1.14.0` is broken when a page contains sub requests, all the files managed by the package run twice. The event `KernelEvents::FINISH_REQUEST` should be handled for MainRequest only.

Sadly we are stuck at 1.13.2 until it's fixed


weaverryan:
Thank you Tarik! This is a complex issue. Let's see if this can fix the bad behavior without introducing any other problems.

PR #175, opened 2022-05-18
Log instead of throw when manifest.json is missing

nicolas-grekas:
Now that webapp-pack [contains webpack-encore-bundle](https://github.com/symfony/webapp-pack/blob/main/composer.json#L31), extending [`base.html.twig`](https://github.com/symfony/recipes/blob/main/symfony/twig-bundle/5.4/templates/base.html.twig) throws an exception unless one runs `yarn` beforehand.

This is bad DX for newcomers. Instead, I propose to log a warning. It's not like empty apps have any mandatory JS :)


nicolas-grekas:
:facepalm: Closing in favor of https://github.com/symfony/recipes/pull/1087

PR #178, opened 2022-06-02
Add Stimulus Twig filters, handle action parameters & allow filters/functions to return array

### LINKED ISSUE #159: Render action params in stimulus_action()
Now that stimulus 3 supports action params (https://world.hey.com/hotwired/stimulus-3-c438d432), the twig function stimulus_action() should render them, such as

```
{{ stimulus_action('item', 'click', 'upvote', {id: 12345, url: '/votes', active: true}) }}

<div data-controller="item">
  <button data-action="item#upvote" 
    data-item-id-param="12345" 
    data-item-url-param="/votes"
    data-item-payload-param='{"value":"1234567"}' 
    data-item-active-param="true">…</button>
</div>
```

I think this would be done similarly to how stimulus_controller handles values, with an optional array.

I also find myself sometimes wanting to pass data- attributes to a target, since stimulus supports getting all targets and walking through them.

### LINKED ISSUE #176: [RFC] Allow stimulus_*() Twig functions to return arrays instead of strings
Hi,

In some cases (most being using Stimulus with Forms) it would be useful that `stimulus_*()` Twig functions return an array of attributes instead of strings, so it could be applied directly, eg.:

```twig
{{ form_row(form.type, { 'attr': stimulus_action('my_controller', 'actionName', 'change' }) }}
```

Returning an array would also allow to merge the attributes:

```twig
{{ form_row(form.type, { 'attr': stimulus_action('my-controller', 'actionName', 'change' })|merge({ 'class': 'customClass' }) }}
```

Currently we're supposed to either use `field_*()` functions or write the attributes ourself I guess:

```twig
{{ form_row(form.type, { 'attr': { 'data-action': 'change->my-controller#actionName' } }) }}
```

WDYT?

jmsche:
Hi everyone,

This PR has several goals:

* make `stimulus_action()` handle action parameters
* add Stimulus Twig filters (so you can e.g. pass different parameters to different actions) (closes #159)
* make Stimulus functions/filters return a Stringable DTO...
* so you can retrieve attributes from these functions/filters as an array (using `.toArray()`), which is helpful for forms (closes #176)

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,16 +229,18 @@ Any non-scalar values (like `data: [1, 2, 3, 4]`) are JSON-encoded. And all
 values are properly escaped (the string `&#x5B;` is an escaped
 `[` character, so the attribute is really `[1,2,3,4]`).
 
-If you have multiple controllers on the same element, pass them all as an
-associative array in the first argument:
+If you have multiple controllers on the same element, you can chain them as there's also a `stimulus_controller` filter:
 
 ```twig
-<div {{ stimulus_controller({
-    'chart': { 'name': 'Likes' },
-    'other-controller': { },
-) }}>
+<div {{ stimulus_controller('chart', { 'name': 'Likes' })|stimulus_controller('other-controller') }}>
     Hello
 </div>
+```
+
+You can also retrieve the generated attributes as an array, which can be helpful e.g. for forms:
+
+```twig
+{{ form_start(form, { attr: stimulus_controller('chart', { 'name': 'Likes' }).toArray() }) }}
 ```
 
 ### stimulus_action
@@ -256,21 +258,33 @@ For example:
 <div data-action="click->controller#method">Hello</div>
 ```
 
-If you have multiple actions and/or methods on the same element, pass them all as an
-associative array in the first argument:
+If you have multiple actions and/or methods on the same element, you can chain them as there's also a
+`stimulus_action` filter:
 
 ```twig
-<div {{ stimulus_action({
- 'controller': 'method',
- 'other-controller': ['method', {'resize@window': 'onWindowResize'}]
-}) }}>
+<div {{ stimulus_action('controller', 'method')|stimulus_action('other-controller', 'test') }}>
     Hello
 </div>
 
 <!-- would render -->
-<div data-action="controller#method other-controller#method resize@window->other-controller#onWindowResize">
+<div data-action="controller#method other-controller#test">
     Hello
 </div>
+```
+
+You can also retrieve the generated attributes as an array, which can be helpful e.g. for forms:
+
+```twig
+{{ form_row(form.password, { attr: stimulus_action('hello-controller', 'checkPasswordStrength').toArray() }) }}
+```
+
+You can also pass [parameters](https://stimulus.hotwired.dev/reference/actions#action-parameters) to actions:
+
+```twig
+<div {{ stimulus_action('hello-controller', 'method', 'click', { 'count': 3 }) }}>Hello</div>
+
+<!-- would render -->
+<div data-action="click->hello-controller#method" data-hello-controller-count-param="3">Hello</div>
 ```
 
 ### stimulus_target
@@ -288,14 +302,10 @@ For example:
 <div data-controller-target="a-target second-target">Hello</div>
 ```
 
-If you have multiple targets on the same element, pass them all as an
-associative array in the first argument:
+If you have multiple targets on the same element, you can chain them as there's also a `stimulus_target` filter:
 
 ```twig
-<div {{ stimulus_target({
- 'controller': 'a-target',
- 'other-controller': 'another-target'
-}) }}>
+<div {{ stimulus_target('controller', 'a-target')|stimulus_target('other-controller', 'another-target') }}>
     Hello
 </div>
 
@@ -303,6 +313,12 @@ associative array in the first argument:
 <div data-controller-target="a-target" data-other-controller-target="another-target">
     Hello
 </div>
+```
+
+You can also retrieve the generated attributes as an array, which can be helpful e.g. for forms:
+
+```twig
+{{ form_row(form.password, { attr: stimulus_target('hello-controller', 'a-target').toArray() }) }}
 ```
 
 Ok, have fun!
--- a/src/Dto/AbstractStimulusDto.php
+++ b/src/Dto/AbstractStimulusDto.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Symfony WebpackEncoreBundle package.
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\WebpackEncoreBundle\Dto;
+
+use Twig\Environment;
+
+/**
+ * @internal
+ */
+abstract class AbstractStimulusDto implements \Stringable
+{
+    /**
+     * @var Environment
+     */
+    private $env;
+
+    public function __construct(Environment $env)
+    {
+        $this->env = $env;
+    }
+
+    abstract public function toArray(): array;
+
+    protected function getFormattedControllerName(string $controllerName): string
+    {
+        return $this->escapeAsHtmlAttr($this->normalizeControllerName($controllerName));
+    }
+
+    protected function getFormattedValue($value)
+    {
+        if ($value instanceof \Stringable || (\is_object($value) && \is_callable([$value, '__toString']))) {
+            $value = (string) $value;
+        } elseif (!is_scalar($value)) {
+            $value = json_encode($value);
+        } elseif (\is_bool($value)) {
+            $value = $value ? 'true' : 'false';
+        }
+
+        return $this->escapeAsHtmlAttr($value);
+    }
+
+    protected function escapeAsHtmlAttr($value): string
+    {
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -241,6 +241,23 @@ associative array in the first argument:
 </div>
 ```
 
+You can chain calls if you prefer as there's also a `stimulus_controller` filter:
```
weaverryan: Let's go further :). I think we should remove the documentation for the "old" entirely. It's quite ugly and hard to understand. Eventually, we should even deprecate it.

README.md
```diff
@@ -241,6 +241,23 @@ associative array in the first argument:
...
+    Hello
+</div>
+```
+
+You can also retrieve the generated attributes as an array, which can be helpful e.g. for forms:
+
+```twig
+{{ form_start(form, { 'attr': stimulus_controller({
+    'chart': { 'name': 'Likes' },
+    'other-controller': {},
```
weaverryan: You an remove the `'` around attr & chart

weaverryan: And let's show the single-controller example here - multiple controllers is quite rare, but using with forms (with a single controller) is quite common.

README.md
```diff
@@ -273,6 +290,32 @@ associative array in the first argument:
 </div>
 ```
 
+You can also pass parameters to actions:
```
weaverryan: How about we link `parameters` to the page on the Stimulus docs about action parameters.

README.md
```diff
@@ -273,6 +290,32 @@ associative array in the first argument:
...
+
+```twig
+<div {{ stimulus_action('controller', 'method', 'click', { 'count': 3 }) }}>Hello</div>
```
weaverryan:
```suggestion
<div {{ stimulus_action('hello-controller', 'method', 'click', { 'count': 3 }) }}>Hello</div>
```

... just to be clear that the first arg is your controller name, not the specific word `controller`.

README.md
```diff
@@ -273,6 +290,32 @@ associative array in the first argument:
...
+<!-- would render -->
+<div data-action="click->controller#method" data-controller-count-param="3">Hello</div>
+```
+
+You can chain calls if you prefer as there's also a `stimulus_action` filter:
+
+```twig
+<div {{ stimulus_action('controller', 'method', 'click', { 'count': 3 })|stimulus_action('other-controller', 'test') }}>
+    Hello
+</div>
```
weaverryan: Here, just like my comment above, let's remove the old documentation for the "old" way of doing multiple actions. So, an example similar to this would move UP to that section and replace the old stuff.

README.md
```diff
@@ -273,6 +290,32 @@ associative array in the first argument:
...
+```
+
+You can also retrieve the generated attributes as an array, which can be helpful e.g. for forms:
+
+```twig
+{{ form_start(form, { 'attr': stimulus_action({
+    'controller': 'method',
+    'other-controller': ['method', {'resize@window': 'onWindowResize'}]
+).toArray() }) }}
```
weaverryan: Use a single-action example... and let's use a field as an example (e.g. `form_row(form.password)`) as an example, perhaps calling some `checkPasswordStrength` method).

README.md
```diff
@@ -305,4 +348,21 @@ associative array in the first argument:
 </div>
 ```
 
+You can chain calls if you prefer as there's also a `stimulus_target` filter:
+
+```twig
+<div {{ stimulus_target('controller', 'a-target')|stimulus_target('other-controller', 'another-target') }}>
+    Hello
+</div>
```
weaverryan: Same as above: replace the old docs with this new way.

README.md
```diff
@@ -305,4 +348,21 @@ associative array in the first argument:
...
+```
+
+You can also retrieve the generated attributes as an array, which can be helpful e.g. for forms:
+
+```twig
+{{ form_start(form, { 'attr': stimulus_target({
+    'controller': 'a-target',
+    'other-controller': 'another-target'
+).toArray() }) }}
```
weaverryan: Same as above: use a single-controller example... and use this with a field instead of `form_start()` to be a bit more realistic

src/Dto/AbstractStimulusDto.php
```diff
@@ -0,0 +1,77 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\WebpackEncoreBundle\Dto;
+
+use Twig\Environment;
+
+abstract class AbstractStimulusDto implements \Stringable
```
weaverryan, line 19: Let's mark it as `@internal`.

src/Dto/AbstractStimulusDto.php
```diff
@@ -0,0 +1,77 @@
...
+        return $this->escapeAsHtmlAttr($value);
+    }
+
+    /**
+     * Normalize a Stimulus Value API key into its HTML equivalent ("kebab case").
+     * Backport features from symfony/string.
+     *
+     * @see https://stimulus.hotwired.dev/reference/values
+     */
+    protected function normalizeKeyName(string $str): string
```
weaverryan: Unless I'm missing something, I think this is only used in one sub-class, so it could be moved there.

weaverryan (commented):
Ran out of time to review everything, but here is a start. I love it!

jmsche:
All your remarks should be fixed :)

jmsche:
I just checked on a project the changes about DTO, .toArray() & action parameters, seems okay :)

README.md
```diff
@@ -256,23 +258,35 @@ For example:
...
+
+<!-- would render -->
+<div data-action="click->hello-controller#method" data-hello-controller-count-param="3">Hello</div>
+```
+
+You can also retrieve the generated attributes as an array, which can be helpful e.g. for forms:
+
+```twig
+{{ form_row(form.password, { attr: stimulus_action('hello-controller', 'checkPasswordStrength').toArray() }) }}
+```
```
weaverryan: Let's move the `.toArray()` example above the "action parameters" example.

src/Twig/StimulusTwigExtension.php
```diff
@@ -33,50 +63,11 @@ public function getFunctions(): array
...
-
-                $key = twig_escape_filter($env, $this->normalizeKeyName($key), 'html_attr');
-                $value = twig_escape_filter($env, $value, 'html_attr');
-
-                $values[] = 'data-'.$controllerName.'-'.$key.'-value="'.$value.'"';
-            }
-        }
-
-        return rtrim('data-controller="'.implode(' ', $controllers).'" '.implode(' ', $values));
+        $dto->addController($dataOrControllerName, $controllerValues);
```
weaverryan, line 68: Just wondering... could we just have ONE `Dto` object - e.g. `StimulusAttributeDto`? It's definitely an edge-case, but I'm wondering about:

```twig
{{ form_row(form.name, {
    attr: stimulus_action('controller', 'doFoo')|stimulus_target('controller', 'fooTarget')
}) }}
```

It seems like that should be "allowed".

jmsche: I could make that happen, but I guess the following can be written instead:

```twig
{{ form_row(form.name, {
    attr: stimulus_action('controller', 'doFoo').toArray()|merge(stimulus_target('controller', 'fooTarget').toArray())
}) }}
```

Also I'm not sure your example can work as attr expects an array I guess?

README.md
jmsche: Done :)

src/Twig/StimulusTwigExtension.php
jmsche: Hi Ryan, little ping :)

weaverryan:
Thank you @jmsche - this is really great!

jmsche:
Thank you Ryan :)

Kocal:
Really smart solution, wp!

jmsche:
Thanks Kocal :)

jeroennoten:
I think that this feature does double escaping:

This Twig code:
```twig
    {{ form_row(form.setNumber, {
        'attr':  stimulus_controller('symfony/ux-autocomplete/autocomplete', {
            url: path('ux_entity_autocomplete', { alias: 'food' })
        }).toArray()
    }) }}
```
Renders double escaping:

```html
<div class="mb-3">
    <label for="purchase_setNumber" class="form-label required">Set number</label>
    <input type="text" id="purchase_setNumber" name="purchase[setNumber]" required="required" maxlength="7"
           data-controller="symfony--ux-autocomplete--autocomplete"
           data-symfony--ux-autocomplete--autocomplete-url-value="&amp;#x2F;autocomplete&amp;#x2F;food"
           class="form-control"/>
</div>
```

Compare to the 'regular' use:
```twig
    <input
            {{ stimulus_controller('symfony/ux-autocomplete/autocomplete', {
                url: path('ux_entity_autocomplete', { alias: 'food' })
            }) }}
    >
```

Which renders correctly:
```html
<input
  data-controller="symfony--ux-autocomplete--autocomplete" data-symfony--ux-autocomplete--autocomplete-url-value="&#x2F;autocomplete&#x2F;food"
  >
```

weaverryan:
Hmm, is the final value actually messed up then due to the double-escaping?

jeroennoten:
Yes, for example autocomplete does not work because the URL is wrong.

I think that the escaping of the attribute values should be moved to the `__toString()` method. Is that an option?

jeroennoten:
@weaverryan, I fixed this bug in https://github.com/symfony/webpack-encore-bundle/pull/202

PR #179, opened 2022-06-02
Run PHP 7.1 in CI

jmsche:
Lowest PHP version supported by the bundle is PHP 7.1, however it's not tested in CI but all other versions are, so let's see if tests pass with it :)


jmsche:
AFAICT the PHP CS Fixer failure is fixed in #178

weaverryan:
Thanks @jmsche!

PR #181, opened 2022-06-15
Minor improvements

nicolas-grekas:


src/Twig/EntryFilesTwigExtension.php
```diff
@@ -80,4 +81,12 @@ private function getTagRenderer(): TagRenderer
     {
         return $this->container->get('webpack_encore.tag_renderer');
     }
+
+    public static function getSubscribedServices(): array
+    {
+        return [
+            'webpack_encore.entrypoint_lookup_collection' => EntrypointLookupInterface::class,
+            'webpack_encore.tag_renderer' => TagRenderer::class,
+        ];
```
weaverryan, line 90: What's the point of adding this here *and* still having it defined in the service? https://github.com/symfony/webpack-encore-bundle/blob/main/src/Resources/config/services.xml#L29-L35 

Can we have just *this* one (and remove the service config)? And if so, will it work even though there is no service with the id `TagRenderer::class`?

maxhelias (approved):

Kocal:
Status: Needs review

(I'm testing CarsonBot integration)

PR #182, opened 2022-06-15
Disable strict mode by default

nicolas-grekas:
For consistency with the strict mode of the asset component, which is disabled by default.

Related to https://github.com/symfony/recipes/pull/1087


GromNaN:
I think we should push for enabling the strict mode as much as possible; even on prod. This option is made to trigger early errors instead of generating incorrect asset urls that breaks page rendering.

Also, modifying the default config value is a breaking change, right? It should have a migration path with deprecation messages.

nicolas-grekas:
It's not a BC break: existing apps will continue to work just fine.
Please check https://github.com/symfony/symfony/pull/46689 and the linked recipe to see why this is important.

nicolas-grekas:
Closing in favor of https://github.com/symfony/webapp-pack/pull/7

PR #185, opened 2022-07-06
[ci] fix flex plugin & legacy stimulus tests

jrushlow:
- adds `flex` to composers allowed plugins
- moves Stimulus `action`, `controller`, & `target` test cases w/ array's as the first arg to legacy tests. (silence deprecation's in CI)


weaverryan:
Gorgeous - thanks Jesse!

PR #187, opened 2022-07-07
Improve Stimulus phpdoc

jmsche:
As discussed with @weaverryan on Slack, I'm changing the phpdoc of Stimulus related code to include only the "correct" way to handle things.

Hope it will pass Static analysis :crossed_fingers:


weaverryan:
Thanks @jmsche!

PR #189, opened 2022-07-13
Moving deprecated code handling for stimulus_ functions into Twig extension

weaverryan:
Hi!

This follows up on @jmsche continued work with the Stimulus functions. Basically:

A) The only deprecated should be in the Twig extension: we deprecated passing certain args as an array there. And so, the code to "normalize" the deprecated arguments should live there.

B) This allows the DTO classes to be "pure" classes that are not encumbered by any deprecation later.

Cheers!

### DIFF
--- a/src/Asset/EntrypointLookup.php
+++ b/src/Asset/EntrypointLookup.php
@@ -134,7 +134,7 @@ class EntrypointLookup implements EntrypointLookupInterface, IntegrityDataProvid
             throw new \InvalidArgumentException(sprintf('Could not find an "entrypoints" key in the "%s" file', $this->entrypointJsonPath));
         }
 
-        if ($this->cache) {
+        if (isset($cached)) {
             $this->cache->save($cached->set($this->entriesData));
         }
 
--- a/src/Dto/StimulusActionsDto.php
+++ b/src/Dto/StimulusActionsDto.php
@@ -17,57 +17,21 @@ final class StimulusActionsDto extends AbstractStimulusDto
     private $parameters = [];
 
     /**
-     * @param string      $controllerName the Stimulus controller name
-     * @param string      $actionName     the action to trigger
-     * @param string|null $eventName      The event to listen to trigger. Optional.
-     * @param array       $parameters     Parameters to pass to the action. Optional.
-     *
-     * @throws \Twig\Error\RuntimeError
+     * @param array $parameters Parameters to pass to the action. Optional.
      */
-    public function addAction($controllerName, string $actionName = null, string $eventName = null, array $parameters = []): void
+    public function addAction(string $controllerName, string $actionName, string $eventName = null, array $parameters = []): void
     {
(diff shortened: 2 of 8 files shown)


src/Dto/StimulusActionsDto.php
```diff
@@ -17,57 +17,21 @@ final class StimulusActionsDto extends AbstractStimulusDto
...
-     * @param string      $controllerName the Stimulus controller name
-     * @param string      $actionName     the action to trigger
-     * @param string|null $eventName      The event to listen to trigger. Optional.
-     * @param array       $parameters     Parameters to pass to the action. Optional.
-     *
-     * @throws \Twig\Error\RuntimeError
+     * @param array  $parameters Parameters to pass to the action. Optional.
      */
-    public function addAction($controllerName, string $actionName = null, string $eventName = null, array $parameters = []): void
+    public function addAction(string $controllerName, string $actionName, string $eventName = null, array $parameters = []): void
```
weaverryan, line 22: The addition of the `string` argument type (which i also added to the other Dto classes) is technically a BC break if you're using version 1.15.0 exactly. However, to have code that would trigger that break, you would have needed to start using these classes directly in 1.15.0 and pass the array argument... which would have triggered deprecations immediately (you don't see it here because it's already moved, but in 1.15.0 a deprecation is triggered in this method if you pass an array to `$controllerName`). So, I think it's not a problem.

src/Dto/StimulusControllersDto.php
```diff
@@ -16,42 +16,20 @@ final class StimulusControllersDto extends AbstractStimulusDto
...
-                }
+            $key = $this->escapeAsHtmlAttr($this->normalizeKeyName($key));
+            $value = $this->getFormattedValue($value);
 
-                $key = $this->escapeAsHtmlAttr($this->normalizeKeyName($key));
-                $value = $this->getFormattedValue($value);
-
-                $this->values['data-'.$controllerName.'-'.$key.'-value'] = $value;
-            }
+            $this->values['data-'.$controllerName.'-'.$key.'-value'] = $value;
```
weaverryan, line 32: All of this is covered by tests for controllers, actions and targets (I double-checked).

src/Asset/EntrypointLookup.php
```diff
@@ -134,7 +134,7 @@ private function getEntriesData(): array
             throw new \InvalidArgumentException(sprintf('Could not find an "entrypoints" key in the "%s" file', $this->entrypointJsonPath));
         }
 
-        if ($this->cache) {
+        if ($this->cache && isset($cached)) {
             $this->cache->save($cached->set($this->entriesData));
         }
```
jrushlow: This technically could be just `if (isset($cached)` because `$cached` is only set if `$this->cache === true` earlier in the method

jrushlow (approved):
Looks good!

PR #191, opened 2022-07-18
Handle Stimulus CSS Classes

### LINKED ISSUE #123: Add classes support to {{ stimulus_controller }}

jmsche:
Closes #123.

This PR aims to add support for Stimulus CSS Classes (see https://stimulus.hotwired.dev/reference/css-classes).

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,8 @@ class ScriptNonceSubscriber implements EventSubscriberInterface
 ### stimulus_controller
 
 This bundle also ships with a special `stimulus_controller()` Twig function
-that can be used to render [Stimulus Controllers & Values](https://stimulus.hotwired.dev/reference/values).
+that can be used to render [Stimulus Controllers & Values](https://stimulus.hotwired.dev/reference/values)
+and [CSS Classes](https://stimulus.hotwired.dev/reference/css-classes).
 See [stimulus-bridge](https://github.com/symfony/stimulus-bridge) for more details.
 
 For example:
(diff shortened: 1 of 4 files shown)


jmsche:
Hi @weaverryan, would you mind reviewing this PR soon?

I'd love to use this feature in my projects.

Thanks :)

README.md
```diff
@@ -204,13 +204,14 @@ class ScriptNonceSubscriber implements EventSubscriberInterface
...
-that can be used to render [Stimulus Controllers & Values](https://stimulus.hotwired.dev/reference/values).
+that can be used to render [Stimulus Controllers & Values](https://stimulus.hotwired.dev/reference/values)
+and [CSS Classes](https://stimulus.hotwired.dev/reference/css-classes).
 See [stimulus-bridge](https://github.com/symfony/stimulus-bridge) for more details.
 
 For example:
 
 ```twig
-<div {{ stimulus_controller('chart', { 'name': 'Likes', 'data': [1, 2, 3, 4] }) }}>
+<div {{ stimulus_controller('chart', { 'name': 'Likes', 'data': [1, 2, 3, 4] }, { 'loading': 'spinner' }) }}>
```
weaverryan, line 231: I think we should leave this example alone, then add a 2nd example below. The reason is that this might make it look like passing CSS values is required or important, whereas in most cases, it wouldn't be used.

weaverryan (commented):
Looks good - can you just tweak that one part in the docs?

README.md
jmsche: Thanks for the review :)

I moved it into a separate example, with & without values.

jmsche:
Code style should be fixed by #194.

I don't know why Tests are cancelled by GitHub :/ 

[Edit] CI is fixed in #194 as well; ubuntu-18.04 is deprecated, hence the temporary failure.

weaverryan (approved):
Looks good! I'll merge and then we'll make sure tests are passing now that you've fixed them :)

weaverryan:
Thanks @jmsche!

PR #202, opened 2023-01-15
Fix double escaping in stimulus DTOs when using toArray() in combination with form methods

jeroennoten:
This fixes a double-escaping bug when the `toArray()` methods are used in combination with the `form_` functions, as described in the README.

`toArray()` should return an array with non-escaped attribute key-value pairs, because they will be escaped when printed.

Therefore, I moved the escaping of the values to the `toString()` methods and I added some additional tests for the DTO classes to verify the change.

This additionally fixes a missed escape for the `data-[controller]-[key]-class` attribute values.

### DIFF
--- a/src/Dto/AbstractStimulusDto.php
+++ b/src/Dto/AbstractStimulusDto.php
@@ -45,7 +45,7 @@ abstract class AbstractStimulusDto implements \Stringable
             $value = $value ? 'true' : 'false';
         }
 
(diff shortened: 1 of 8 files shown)


weaverryan (approved):
That's really nice. It makes sense: the `__toString()` will give you the escaped version. But `toArray()` will give you the non-escaped version, and then it's up to you to render it correctly.

weaverryan:
Excellent PR Jeroen!

PR #206, opened 2023-02-13
Add Twig function and filter for Stimulus Outlets integration

pbories:
The Outlets API was introduced in Stimulus release [v3.2.0](https://github.com/hotwired/stimulus/releases/tag/v3.2.0).

This PR intends to ease its integration in Twig templates, just like Stimulus controllers, actions and targets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ class ScriptNonceSubscriber implements EventSubscriberInterface
 ### stimulus_controller
 
 This bundle also ships with a special `stimulus_controller()` Twig function
-that can be used to render [Stimulus Controllers & Values](https://stimulus.hotwired.dev/reference/values)
+that can be used to render [Stimulus Controllers & Values](https://stimulus.hotwired.dev/reference/values), [Outlets](https://stimulus.hotwired.dev/reference/outlets)
 and [CSS Classes](https://stimulus.hotwired.dev/reference/css-classes).
 See [stimulus-bridge](https://github.com/symfony/stimulus-bridge) for more details.
 
@@ -258,6 +258,17 @@ If you have multiple controllers on the same element, you can chain them as ther
 <div {{ stimulus_controller('chart', { 'name': 'Likes' })|stimulus_controller('other-controller') }}>
     Hello
 </div>
+```
+
+If you need to attach an [outlet](https://stimulus.hotwired.dev/reference/outlets) to the controller, you can call the `addOutlet()` method.
+
+For example:
+
+```twig
+<div {{ stimulus_controller('chart').addOutlet('outlet-controller', '.css-selector') }}>Hello</div>
+
+<!-- would render -->
+<div data-controller="chart" data-chart-outlet-controller-outlet=".css-selector">Hello</div>
 ```
 
 You can also retrieve the generated attributes as an array, which can be helpful e.g. for forms:
--- a/src/Dto/StimulusControllersDto.php
+++ b/src/Dto/StimulusControllersDto.php
@@ -51,7 +51,7 @@ final class StimulusControllersDto extends AbstractStimulusDto
 
         $this->outlets['data-'.$this->controllers[0].'-'.$outletName.'-outlet'] = $selector;
 
-        return new Markup($this, 'UTF-8');
+        return $this;
     }
 
     public function __toString(): string
@@ -61,7 +61,7 @@ final class StimulusControllersDto extends AbstractStimulusDto
         }
 
         return rtrim(
-            'data-controller="'.implode(' ', $this->controllers).'" '.
+            'data-controller='.implode(' ', $this->controllers).
             implode(' ', array_map(function (string $attribute, string $value): string {
                 return $attribute.'="'.$this->escapeAsHtmlAttr($value).'"';
             }, array_keys($this->values), $this->values)).' '.
@@ -69,7 +69,7 @@ final class StimulusControllersDto extends AbstractStimulusDto
                 return $attribute.'="'.$this->escapeAsHtmlAttr($value).'"';
             }, array_keys($this->classes), $this->classes)).' '.
             implode(' ', array_map(function (string $attribute, string $value): string {
-                return $attribute.'="'.$this->escapeAsHtmlAttr($value).'"';
+                return $attribute.'='.$this->escapeAsHtmlAttr($value);
             }, array_keys($this->outlets), $this->outlets))
         );
     }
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -344,4 +344,36 @@ You can also retrieve the generated attributes as an array, which can be helpful
...
 ```
 
+### stimulus_outlet
+
+The `stimulus_outlet()` Twig function can be used to render [Stimulus Outlets](https://stimulus.hotwired.dev/reference/outlets).
+
+For example:
+
+```twig
+<div {{ stimulus_outlet('controller', 'foo', '.selector') }}>Hello</div>
```
weaverryan: I haven't worked with outlets yet (I've only read the docs). If I'm correct, an outlet would always be on the same element as a controller, correct? So something like:

```
<div {{ stimulus_controller('hello') }} {{ stimulus_outlet('hello', 'foo', '.selector') }}>Hello</div<
```

Is that correct? If so, we should try to "fit" its syntax into `stimulus_controller` somehow - it feels unnecessary to require 2x `stimulus_` on the same element. For example, perhaps:

```
<div {{ stimulus_controller('hello')|stimulus_outlet('foo', '.selector')>
```

WDYT? Probably it would only make sense to make `stimulus_outlet` a filter - never its own function that could be used standalone. Or perhaps you can think of a better syntax. When `stimulus_controller` was created, CSS classes & outlets didn't exist... so it's getting kind of busy :p. For example, maybe this is better?

```
<div {{ stimulus_controller('hello').addOutlet('foo', '.selector')>
```

pbories: Thank you @weaverryan for your answer (I'm a big fan of SymfonyCasts) !
You're right about the doc, I should have mentioned that an outlet would always be on the same element as a controller.
IMHO, a filter would not be the best thing to do because it would only work when coupled with a controller.
So you're right again : let me try to create a new function in StimulusControllersDto !

pbories: Tell me if I'm wrong @weaverryan, but isn't there a mistake in StimulusControllersDto, line 39 ?

`$this->values['data-'.$controllerName.'-'.$key.'-class'] = $class;`

I guess it should be `$this->classes['...`, right ?

src/Dto/StimulusControllersDto.php
```diff
@@ -40,20 +43,34 @@ public function addController(string $controllerName, array $controllerValues =
...
+
     public function __toString(): string
     {
         if (0 === \count($this->controllers)) {
             return '';
         }
 
         return rtrim(
-            'data-controller="'.implode(' ', $this->controllers).'" '.
+            'data-controller='.implode(' ', $this->controllers).
```
weaverryan: We do want the `"` still

src/Dto/StimulusControllersDto.php
```diff
@@ -40,20 +43,34 @@ public function addController(string $controllerName, array $controllerValues =
...
             implode(' ', array_map(function (string $attribute, string $value): string {
                 return $attribute.'="'.$this->escapeAsHtmlAttr($value).'"';
             }, array_keys($this->values), $this->values)).' '.
             implode(' ', array_map(function (string $attribute, string $value): string {
                 return $attribute.'="'.$this->escapeAsHtmlAttr($value).'"';
-            }, array_keys($this->classes), $this->classes))
+            }, array_keys($this->classes), $this->classes)).' '.
+            implode(' ', array_map(function (string $attribute, string $value): string {
+                return $attribute.'='.$this->escapeAsHtmlAttr($value);
+            }, array_keys($this->outlets), $this->outlets))
```
weaverryan, line 73: Can you add a test for this - in `StimulusControllersDtoTest`?

README.md
```diff
@@ -260,6 +260,17 @@ If you have multiple controllers on the same element, you can chain them as ther
 </div>
 ```
 
+If you need to attach an outlet to some controller, you can call the addOutlet() method.
```
weaverryan:
```suggestion
If you need to attach an [outlet](https://stimulus.hotwired.dev/reference/outlets)
to the controller, you can call the `addOutlet()` method.
```

README.md
```diff
@@ -260,6 +260,17 @@ If you have multiple controllers on the same element, you can chain them as ther
...
+
+For example:
+
+```twig
+<div {{ stimulus_controller('foo').addOutlet('bar', '.bar') }}>Hello</div>
```
weaverryan:
```suggestion
<div {{ stimulus_controller('chart').addOutlet('outlet-controller', '.css-selector') }}>Hello</div>
```

src/Dto/StimulusControllersDto.php
```diff
@@ -11,10 +11,13 @@
 
 namespace Symfony\WebpackEncoreBundle\Dto;
 
+use Twig\Markup;
```
weaverryan, line 14: Not needed

README.md
weaverryan: Woh, yea, totally! Can you please fix it? It actually doesn't break anything, but we shouldn't leave it that way :)

src/Dto/StimulusControllersDto.php
```diff
@@ -36,10 +39,21 @@ public function addController(string $controllerName, array $controllerValues =
...
 
+    public function addOutlet(string $outletName, string $selector)
+    {
+        if (1 < \count($this->controllers)) {
+            throw new \LengthException('You cannot call addOutlet() method when passing more than one controller identifier to stimulus_controller() function');
+        }
+
+        $this->outlets['data-'.$this->controllers[0].'-'.$outletName.'-outlet'] = $selector;
+
+        return new Markup($this, 'UTF-8');
```
weaverryan, line 54: Why is the `Markup` class suddenly needed in this situation?

pbories: I'm probably doing something wrong but if I don't use this class, the result becomes : `data-controller=""controller"" data-controller-foo-outlet="".foo""`

weaverryan: Weird. It make me think that, below in `__toString()`, when we call `$this->escapeAsHtmlAttr($value)`, that is taking `.foo` and turning it into `".foo"`... but that seems very odd...

pbories: It's also clean when I use the raw filter like this : {{ stimulus_controller('foo').addOutlet('bar', '.bar')|raw }}, but it's certainly not what we need.

weaverryan:
Outlets added to StimulusBundle! Use the functions from that bundle instead. Thanks @pbories for getting this rolling

Cheers!

PR #211, opened 2023-05-17
[WIP] Deprecating the stimulus functionality in favor of an upcoming StimulusBundle

weaverryan:
Hi!

This isn't ready yet, but a new StimulusBundle is being created that will now be the home of the `stimulus_` Twig functions 🎆 . This PR is to deprecate those functions and only load them if the StimulusBundle is not present.

Cheers!


weaverryan:
~I'm going to keep this open, but delay merging it for a bit - as StimulusBundle will initially start as experimental, just to be safe. I don't want to put in the deprecation and force people to switch to an experimental library.~

Instead, in StimulusBundle - https://github.com/symfony/ux/pull/887 - I'll set its Twig extension priority to a higher priority so that its`stimulus_` functions take priority over the ones from WebpackEncoreBundle. The idea is that, once you install StimulusBundle, you're opting into the new `stimulus_` functions.

For the most part, the `stimulus_` functions between StimulusBundle & WebpackEncoreBundle are identical. However, internally, they use different objets to accomplish things. So if you're doing something more advanced, you will notice that difference as soon as StimulusBundle is installed.

Update: on 3rd thought, to keep everything clean, we'll merge and tag this soon, and tag a 2.0 without this. That will also help us have separate recipes - the old "style" and the new style.

PR #221, opened 2023-07-24
Allow Symfony 7

jmsche:


weaverryan:
When was that released? :p Thanks @jmsche!

jmsche:
Well, tests pass, so... :smile:

PR #233, opened 2024-08-24
Ensure PHP 8.3 and 8.4 are supported

Kocal:


Kocal:
Will fix PHP-CS-Fixer checks on another PR.

PR #236, opened 2024-10-01
Allow entrypoints.json to be hosted remotely

### LINKED ISSUE #76: Allow output_path to be remote url

Kocal:
Fixes https://github.com/symfony/webpack-encore-bundle/issues/76

See #97, this PR uses the Symfony HttpClient to remotely fetch the entrypoints file, but also to allow mocking in a proper way.
It also add/adapt many tests, and improve the documentation.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
     },
     "require-dev": {
         "symfony/framework-bundle": "^5.4 || ^6.2 || ^7.0",
+        "symfony/http-client": "^5.4 || ^6.2 || ^7.0",
         "symfony/phpunit-bridge": "^5.4 || ^6.2 || ^7.0",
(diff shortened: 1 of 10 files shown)


Kocal:
Thank you Raphaël Louvradoux.

Kocal:
Thank you @rlvdx.

PR #248, opened 2025-08-05
Add support for Symfony 8

Kocal:


xabbuh:
We probably need to tweak the CI matrix a bit as we currently don't have a job running against Symfony 7 (the same was actually already true before for Symfony 6).

Kocal:
Yes! I plan to re-use the matrix from UX Packages (or my packages, based on this same matrix), but with this matrix I'm failing to install Symfony 8 packages (e.g. in https://github.com/Kocal/BiomeJsBundle/pull/33), but here it's working https://github.com/symfony/webpack-encore-bundle/actions/runs/16742430575/job/47393383777?pr=248

Maybe because of `SYMFONY_REQUIRE` env var? 🤔

Kocal:
Nvm, https://github.com/symfony/webpack-encore-bundle/actions/runs/16742766500/job/47394353881?pr=248 correctly install Symfony 8.0.x-dev dependencies, the only difference I see with UX and my packages is how Flex is installed/enabled

.github/workflows/ci.yml
```diff
@@ -8,19 +8,36 @@ on:
...
-    name: "Tests ${{ matrix.php-version }} deps ${{ matrix.dependency-versions }}"
     runs-on: ubuntu-22.04
 
     strategy:
       fail-fast: false
       matrix:
-        # normal, highest, non-dev installs
         php-version: ['8.1', '8.2', '8.3', '8.4']
-        dependency-versions: ['highest']
+        dependency-version: [ '' ]
```
xabbuh:
```suggestion
        dependency-version: ''
```

.github/workflows/ci.yml
```diff
@@ -8,19 +8,36 @@ on:
...
+        symfony-version: [ '' ]
```
xabbuh, line 18:
```suggestion
        symfony-version: ''
```

.github/workflows/ci.yml
Kocal: I think matrix properties only allow array?

.github/workflows/ci.yml
```diff
@@ -8,19 +8,36 @@ on:
...
+          -   php-version: '8.1'
+              dependency-version: 'lowest'
+          # LTS version of Symfony
+          -   php-version: '8.1'
+              symfony-version: '6.4.*'
+          # Explicit Symfony versions
+          -  php-version: '8.1'
+             symfony-version: '5.4'
+          -  php-version: '8.1'
+             symfony-version: '6.2'
```
xabbuh, line 34: I wouldn't waste resources on EOL versions

.github/workflows/ci.yml
```diff
@@ -8,19 +8,36 @@ on:
...
+          -  php-version: '8.2'
+             symfony-version: '7.0'
```
xabbuh, line 36: same here but we can use 7.3 instead

.github/workflows/ci.yml
xabbuh: Oh, is it? Then please ignore my comment.

.github/workflows/ci.yml
Kocal: I specified 6.2 and 7.0 because the Bundle support these versions. I also tend to not really care about unmaintained versions, but is it fine here?

xabbuh: maybe it's about time to drop support for unmaintained Symfony releases

Kocal: I'm totally fine with that, but I was thinking doing that in a second PR, and then release WebpackEncoreBundle 3.0

PR #253, opened 2025-11-27
Tests on Symfony 8 stable

Kocal:


composer.json
```diff
@@ -35,6 +35,11 @@
             "Symfony\\WebpackEncoreBundle\\Tests\\": "tests/"
         }
     },
+    "config": {
+        "audit": {
+            "ignore": ["PKSA-4k7v-pfvw-nqvp"]
```
xabbuh: Why do we need this?

Kocal: Because of https://github.com/symfony/webpack-encore-bundle/actions/runs/19734496862/job/56543072752?pr=253

xabbuh: Looking at the failures I suggest that we remove the jobs for not maintained Symfony versions instead.

Kocal: This is something we already evoked in https://github.com/symfony/webpack-encore-bundle/pull/248#discussion_r2253340999 

... does increasing `^7.0` to `^7.3.7` (the lowest 7.x.y versions that contains the security fix) is something that we can do in a non major release?

Kocal: Ah, no, you mean in the CI, drop testing 7.0 (which is not maintained since long time ago)

let's go for it

.github/workflows/ci.yml
```diff
@@ -30,13 +30,10 @@ jobs:
           # Explicit Symfony versions
           -  php-version: '8.1'
              symfony-version: '5.4.*'
-          -  php-version: '8.1'
-             symfony-version: '6.2.*'
```
xabbuh, line 34: What about using `6.4.*` instead (I think we otherwise don't have a job using this LTS)?

xabbuh: never mind

composer.json
```diff
@@ -35,6 +35,13 @@
             "Symfony\\WebpackEncoreBundle\\Tests\\": "tests/"
         }
     },
+    "config": {
```
xabbuh: can be reverted now

.github/workflows/ci.yml
Kocal: 6.4.* is already tested above

composer.json
Kocal: Yes, done!

xabbuh (approved):

PR #265, opened 2026-07-21
Fall back to the raw path when strict mode rejects an already-resolved entrypoint

### LINKED ISSUE #230: encore_entry_link_tags incorrectly querying manifest.json
I have encountered this while hosting a Symfony app in a subfolder and managed to reproduce it with a minimal test case.

The behavior only shows when `framework.assets.strict_mode` is set to `true`.

## Error

The error thrown is: `Asset "/web-subfolder/app.b75294ae.css" not found in manifest ".../build-output/manifest.json". Did you mean one of these? "my-prefix/app.css", "my-prefix/app.js"`

## Reproduce

To demonstrate the problem I am setting three different values for these three directories:
- output (`setOutputPath`): `'build-output/'`
- web root (`setPublicPath`): `'/web-subfolder'`
- manifest key prefix (`setManifestKeyPrefix`): `'my-prefix/'`

`manifest.json`:

```json
{
  "my-prefix/app.css": "/web-subfolder/app.b75294ae.css",
  "my-prefix/app.js": "/web-subfolder/app.53d71f57.js"
}
```

`entrypoints.json`:

```json
{
  "entrypoints": {
    "app": {
      "js": [
        "/web-subfolder/app.53d71f57.js"
      ],
      "css": [
        "/web-subfolder/app.b75294ae.css"
      ]
    }
  }
}
```

Template:

```twig
{{ encore_entry_link_tags('app') }}
```

## Fix

I'm not really sure I completely understand how all the parts work together, but it seems to me that there is no need to call `getAssetPath` in `TagRenderer->renderWebpackLinkTags` ([here](https://github.com/symfony/webpack-encore-bundle/blob/v2.1.1/src/Asset/TagRenderer.php#L96)), since the entrypoint already provides the "mapped" values.

## Demo

I have created a demo repository [here](https://github.com/bmorg/encore_entry_link_tags_bug). The relevant changes to reproduce are in [this commit](https://github.com/bmorg/encore_entry_link_tags_bug/commit/f95c8a2cb67858b7df9311440287ae8ec2baf87d).

Amoifr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Issues        | Fix #230
| License       | MIT

`entrypoints.json` already stores the final public paths of the built assets, but `TagRenderer` passes each one back through `packages->getUrl()`. When the default asset package is backed by the JSON manifest (`framework.assets.json_manifest_path`) with `framework.assets.strict_mode: true`, the strategy treats the already-resolved path as a manifest key, fails to find it and throws `AssetNotFoundException`, so `encore_entry_link_tags()` / `encore_entry_script_tags()` blow up. With `strict_mode: false` the same call silently returns the path unchanged, which is why the bug only shows up in strict mode (reported by @bmorg, kept open by @Kocal).

I reproduced it against the real asset component: `getUrl('/web-subfolder/app.b75294ae.css')` throws in strict mode, while `getUrl('my-prefix/app.css')` (the manifest key) resolves fine.

This catches `AssetNotFoundException` in `getAssetPath()` and falls back to the raw path, which is already the correct, final one, matching what `strict_mode: false` does today. Real `asset()` calls elsewhere keep their strict-mode behavior untouched. I deliberately kept the `getUrl()` call rather than removing it (as the issue suggested), since it is what applies a base path / CDN to the entrypoint values.

@Kocal, since this touches the entrypoints / manifest / base-path interaction, I went for the smallest safe fix but I'm happy to adjust the direction if you had something else in mind.

Thanks a lot for the bundle! 🙏

### DIFF
--- a/src/Asset/TagRenderer.php
+++ b/src/Asset/TagRenderer.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\WebpackEncoreBundle\Asset;
 
+use Symfony\Component\Asset\Exception\AssetNotFoundException;
 use Symfony\Component\Asset\Packages;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 use Symfony\Contracts\Service\ResetInterface;
@@ -173,10 +174,19 @@ class TagRenderer implements ResetInterface
             throw new \Exception('To render the script or link tags, run "composer require symfony/asset".');
         }
 
-        return $this->packages->getUrl(
-            $assetPath,
-            $packageName
-        );
+        try {
+            return $this->packages->getUrl(
+                $assetPath,
+                $packageName
+            );
+        } catch (AssetNotFoundException) {
+            // entrypoints.json already stores the final public paths, so an
+            // asset package backed by the JSON manifest (with strict_mode
+            // enabled) cannot resolve them a second time and throws. The path
+            // is already the correct one in that case, so use it as-is, which
+            // is also what happens when strict_mode is disabled.
+            return $assetPath;
+        }
     }
 
     private function getEntrypointLookup(string $buildName): EntrypointLookupInterface
--- a/tests/Asset/TagRendererStrictModeTest.php
+++ b/tests/Asset/TagRendererStrictModeTest.php
@@ -1,0 +1,77 @@
+<?php
+
+/*
+ * This file is part of the Symfony WebpackEncoreBundle package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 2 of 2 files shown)


Amoifr:
Heads-up on the red Fabbot check: it is a false positive coming from two of its steps disagreeing with each other. `Check PHP code style` wants a `: void` return type on the test methods, while `Check test-case methods` forbids it ("Test case methods should not have a return type"), so both cannot be satisfied at once.

I went without the return type, which matches every existing test method in the suite and the `Check test-case methods` convention. All the blocking checks (tests on 8.1 to 8.5, PHPStan, Psalm, Composer Normalize) are green. Happy to flip it the other way if you prefer. 🍒

